PR #15, opened 2018-08-15
fixing tests

katiebreivik:


katiebreivik:
This is coming in off of a weird branch because I don't do GitHub very well sometimes

PR #58, opened 2018-11-05
Needed to make a change to the BSE ini file convert to dict

scottcoughlin2014:


katiebreivik:
Looks like you can edit the code from the web. Github is pretty dope.

Issue #80, opened 2018-12-04
Set of initial conditions that do not elicit a final BCM entry


scottcoughlin2014:
I cannot attach the HDF5 file but discuss about this should go here. This is even after the `kstar1` and `kstar2` being 15 hack.

scottcoughlin2014:
I think #82  finally took care of this

PR #81, opened 2018-12-05
added in stopgap to set final bcm array step at very end of evolv2

katiebreivik:


katiebreivik:
This is a PR that should fix the problems where Evolve does not return a final state. see https://github.com/COSMIC-PopSynth/COSMIC/issues/80

Issue #88, opened 2018-12-11
add merger_type series to bcm array


michaelzevin:
Create a merger_type column in the bcm array. This should be set to a default value of -1. The merger type will be specified as a combination of the kstar types (always with a minimum of 2 digits for each kstar type). 

For example, a BH-BH merger will have a merger_type value of 1414, a helium white dwarf with a giant branch naked helium star will be 1009, and a deeply or fully convective low mass MS star with a main sequence star will be 0001.

scottcoughlin2014:
It would be nice to have more examples to test this. But here is a start. Also this new methodlogies means that `1413` get captures correctly where as before I was setting `binstate=14` for those with previous method. Thanks to @michaelzevin for the pointer
```
(cosmic-py36) [sbc538@quser11 debug]$ bash run_all_merger_systems.sh 
          tphys  kstar_1    mass0_1     mass_1  lumin_1   ...     formation_1  formation_2  bin_state  merger_type  bin_num
0  10114.768555     14.0  13.792748  25.247549    -10.0   ...             4.0          4.0        1.0       1414.0        0

[1 rows x 39 columns]
        tphys  kstar_1   mass0_1  mass_1   lumin_1   ...     formation_1  formation_2  bin_state  merger_type  bin_num
0  10118.8125     15.0  1.429424     0.0 -8.373529   ...             4.0          5.0        1.0       1313.0        0

[1 rows x 39 columns]
         tphys  kstar_1   mass0_1    mass_1  lumin_1   ...     formation_1  formation_2  bin_state  merger_type  bin_num
0  3288.084717     14.0  7.805439  4.697084    -10.0   ...             4.0          4.0        1.0       1314.0        0

[1 rows x 39 columns]
         tphys  kstar_1     mass0_1      mass_1    lumin_1   ...     formation_1  formation_2  bin_state  merger_type  bin_num
0     0.068922      1.0  112.169884  112.169884   6.201618   ...             0.0          0.0        0.0         -1.0        0
0  1175.699097     15.0    9.085922    0.000000 -10.000000   ...             4.0          4.0        1.0       1413.0        0
```

scottcoughlin2014:
@katiebreivik Perhaps you could email me a file that would include mergers with `kstar1=9` and `kstar2=9` or something to test that those get set to `0909` correctly.

scottcoughlin2014:
#89 addresses this

Issue #87, opened 2018-12-11
binary_state series in bcm array


michaelzevin:
Create a binary_state series in the bcm array, which takes values of 0 if the system is alive today, 1 if the system has mergerd, and 2 if the system was disrupted.

scottcoughlin2014:
#89 addresses this

PR #63, opened 2018-11-06
WIP: First pass at handling filtering through the ini file. 

scottcoughlin2014:
I have not tackled convergence at all except there is a line in the inifile.

PR #89, opened 2018-12-11
WIP: Update bin state

scottcoughlin2014:
@katiebreivik this is a pretty good start to address the new logic mentioned in #88 #87 

Things to check I think is that we are really capturing all the times the system gets disrupted, from all my mucking about and changing of `bin_state` before separating `bin_state` and `merger_type` I am not 100 percent sure. Also I believe when `bin_state = 2`  gets set I have reset `mergertype=-1` in all those situations.


scottcoughlin2014:
@katiebreivik I need update the unit tests which will for sure fail.

Scotty

katiebreivik:
Merging this now since tests are broken in PR https://github.com/COSMIC-PopSynth/COSMIC/pull/63 and it makes sense to maybe fix them over there

PR #107, opened 2019-01-25
Convergence

katiebreivik:
This pull request fixes the convergence issue we were running into. The problem was two fold:
1 - If there are a small number of binaries (<100) the convergence isn't a super useful test
2 - The convergence was computed by comparing a large population (~10,000) to a very small population (sometimes <5), so it was never converging. What should be compared is the large population containing every binary and a large population containing every binary _except_ the last added population. 

Now, the convergence is only computed if there are at least 100 binaries in the population and continues to simulate new populations until that limit is reached. It is also computed the way 2 above says it should be.

Finally, I realized the orbital period sample was (very rarely, but still happened) initializing binaries with orbital periods that cause the binaries to be in contact before they even started evolving. Now the initial orbital periods are chosen such that the radius of each star (which is assumed to be a MS star) is less than half the star's Roche radius. This is consistent with StarTrack a la Dominik+ (2012, 2013).

### DIFF
--- a/bin/runFixedPop
+++ b/bin/runFixedPop
@@ -292,7 +292,7 @@ if __name__ == '__main__':
                                                   kstar1_range, kstar2_range, log_file)
  
             if len(match_all) > 1:
-                match_all = np.reshape(match_all, (-1,len(filters['binary_state'])))
+                match_all = np.reshape(match_all, (-1,len(args.convergence_params)))
                 match_save = pd.DataFrame(np.atleast_2d(match_all), columns = args.convergence_params)
                 dat_store.append('match', match_save)
             # save the total_sampled_mass so far


scottcoughlin2014:
Thanks Katie I will give this a look in a sec

katiebreivik:
@scottcoughlin2014 wait before you accept this; I'm working on one more update to make sure the filtering is done properly. The way it is now, we can't handle multiple bin_states in a single runFixedPop call since the filtering overwrites the bcm array.

scottcoughlin2014:
@katiebreivik alright no worries, I am going to test the new convergence criteria on some of those `bin_state==1` runs I was doing before and see what it looks like now.

katiebreivik:
Ok! Now I think this is something that filters properly and has worked on a small scale BBH test I did on my laptop.

scottcoughlin2014:
`bin_state == 1` configurations the `bpp_save` seems wrong
```
(Pdb) bcm_conv_1.bin_num.unique().size
120
(Pdb) bcm_conv_2.bin_num.unique().size
60
(Pdb) bpp_save.bin_num.unique().size
28
```

katiebreivik:
@scottcoughlin2014 here is an initial condition for a disrupted BBH:

single_binary = InitialBinaryTable.SingleBinary(m1=27.762951, m2=12.557175, 
                                                porb=470.994687, ecc=0.467888, 
                                                tphysf=12523.076840, kstar1=1, kstar2=1, 
                                                metallicity=0.000288)

cosmic/utils.py
```diff
@@ -61,19 +62,45 @@ def filter_bpp_bcm(bcm, bpp, method):
...
+                                (bpp.kstar_2.isin([10,11,12])) &
+                                (bpp.evol_type == 3.0)].bin_num.tolist()
+            bcm = bcm.loc[~bcm.bin_num.isin(idx_save)]
         elif (meth == 'select_final_state') and use:
             bcm = bcm.iloc[bcm.reset_index().groupby('bin_num').tphys.idxmax()]
         elif (meth == 'binary_state'):
             bcm = bcm.loc[bcm.bin_state.isin(use)]
+            
+            # CREATE A LIST OF BIN_NUM INDICES TO APPEND TO
+            bin_num_save = []
```
scottcoughlin2014: `bin_num_save` looks great but it makes me think should we not have a grand `bin_num_save` list at the beginning of the filtering before we loop over all the methods and simply append to this list the `idx` for all binaries we are saving? so you have `idx_save` for `mass transfer....` and for `LISA_sources` etc. would it be smoother and easier to add new filters if we had a master `bin_num_save` list to append to instead of just in `binary_state`?

bin/runFixedPop
```diff
@@ -256,42 +260,50 @@ if __name__ == '__main__':
...
+            initC_filter = initCond.loc[initCond.bin_num.isin(bcm_filter.bin_num)]
+            bpp_filter = bpp.loc[bpp.bin_num.isin(bcm_filter.bin_num)]
+
         # Run the match on short period binaries, such that we get a better sampling
         # for the highest signal systems
-        if (len(bcm_save_filtered) > 0):
+        n_bin_state = []
+        for bin_state in filters['binary_state']:
+            n_bin_state.append(len(bcm_filter.loc[bcm_filter.bin_state == bin_state]))
+        if all(i >= 50 for i in n_bin_state):
```
scottcoughlin2014: I understand making sure that each binary state of interest has enough samples to be worth testing convergence on. However I wonder about the fact that this logic throws out iterations where we have have 49 systems that made it to bin state 0 1 and 2 each and the code says oh that's not enough and does not save them to the overall `bcm_save` variable, should we not still save them to the array but not calculate convergence? or am I missing something.

bin/runFixedPop
```diff
@@ -256,42 +260,50 @@ if __name__ == '__main__':
...
+            dat_store.append('initCond', initC_filter)
 
             # perform the convergence
             match_all = Match.perform_convergence(args.convergence_params, filters['binary_state'],\
-                                                  convergence, bcm_save, bcm_save_filtered, bpp_save,\
+                                                  convergence, bcm_save, bcm_filter, bpp_save,\
                                                   kstar1_range, kstar2_range, log_file)
+ 
             if len(match_all) > 1:
+                match_all = np.reshape(match_all, (-1,len(filters['binary_state'])))
```
scottcoughlin2014: I *think* based on my testing of only one bin_state we want this to be `len(args.convergence_params)`

katiebreivik:
> One last comment on top of the comments I made in the code, did we determine we were going to adding in a minimum `bin_width` and if so what did we decide? I may have missed it in this PR.
> 
> Scotty

I realized that we don't need to add the minimum `bin_width`; the knuth method of selecting the binwidth works just fine as long as the convergence is being computed properly (which it wasn't initially).

I'm going through the code comments now.

bin/runFixedPop
katiebreivik: @scottcoughlin2014 
I can't figure out how to quote the code, but starting on line 263 of [runFixedPop](https://github.com/COSMIC-PopSynth/COSMIC/pull/107/files#diff-30a06cad5f4b8c2bdc7fcbbad5e6305aR263) is where the `bcm_filter`, `initC_filter` and `bpp_filter` DataFrames are appended to until each `bin_state` has >50 values in it. Once this condition is fulfilled you then compute the convergence, `bcm_save` and `bpp_save` are appended to,  and `bcm_filter`, `initC_filter` and `bpp_filter` are emptied to be filled again.

cosmic/utils.py
katiebreivik: I generally agree with this but am having a difficult time finding the right logic specifically for the mass transfer filtering and the select final state filtering.

The mass transfer filter is difficult because the logic is to throw out values based on some criteria, instead of retaining values based on that criteria. If we just do in the inverse, keep everything that _doesn't_ satisfy the criteria, then we will have to filter it out later on. 

With the select final state filter, we have to appeal to iloc to get the actual highest index in the bcm array to select the final time.

These complications lead me to think that it isn't just as simple as appending to a `bin_num_save` list for all the different types of filtering.

bin/runFixedPop
katiebreivik: You are right: This is fixed [here](https://github.com/COSMIC-PopSynth/COSMIC/pull/107/commits/52a7e8400424f00caf7438fe04ede2906420e80a#diff-30a06cad5f4b8c2bdc7fcbbad5e6305aR295).

Issue #108, opened 2019-01-29
Adding kick magnitudes of Supernovas


scottcoughlin2014:
This will add two columns to the bpp array called `kick_magnitude_1` and `kick_magnitude_2`.

The logic will be as follows, when there is SN for a kick to be recorded than the relevant column(s) will have a value and at all other times will be zero. So for instance is their is a SN of particle 1 and the system survives and if the second particle goes through a SN and the system again survives it would look like

```
          tphys     mass_1     mass_2  kstar_1  kstar_2          sep           ecc        RROL_1    RROL_2  evol_type  kick_magnitude_1  kick_magnitude_2  bin_num
0      4.366515  48.654041  41.406265      2.0      1.0  1428.131348  3.757214e-01  5.656106e-02  0.043885        2.0          0.000000          0.000000      0.0
0      4.369806  48.393070  41.425571      2.0      1.0  1429.464111  3.751530e-01  1.000569e+00  0.043911        3.0          0.000000          0.000000      0.0
0      4.371137  48.175579  41.479919      4.0      1.0  1405.657837  3.635945e-01  3.262424e+00  0.044632        2.0          0.000000          0.000000      0.0
0      4.404452  35.130230  50.676319      4.0      1.0  1195.280273  0.000000e+00  4.915884e+00  0.038660       14.0          0.000000          0.000000      0.0
0      4.439046  19.250885  63.908356      4.0      1.0  1694.894775  0.000000e+00  2.584105e-01  0.022522        4.0          0.000000          0.000000      0.0
0      4.439807  19.136734  63.929607      7.0      1.0  1695.558472  0.000000e+00  2.976952e-03  0.022500        2.0          0.000000          0.000000      0.0
0      4.847520  13.792748  62.638836      8.0      1.0  1844.470093  0.000000e+00  2.393120e-03  0.023013        2.0          0.000000          0.000000      0.0
0      4.858730  11.182409  62.601822     14.0      1.0  1912.809448  3.358260e-02  1.016000e-07  0.021552        2.0          2.444581          0.000000      0.0
0      5.881090  11.182409  59.622791     14.0      2.0  1992.374023  3.358260e-02  9.622744e-08  0.038969        2.0          0.000000          0.000000      0.0
0      5.881470  11.182409  59.621807     14.0      2.0  1992.401611  3.358260e-02  9.622566e-08  0.057872       12.0          0.000000          0.000000      0.0
[...]

PR #109, opened 2019-01-29
Recording black hole kick

scottcoughlin2014:
This addresses #108 in addition it adds more units tests (merged BNS and BBH and NSBH) to go along with the two new columns. By no means is this enough units tests but it is a start.


scottcoughlin2014:
@michaelzevin @katiebreivik alright this latest commit *should* compile and pass the tests (I have added new unit tests that produce NSNS NSBH and BHBH although this is a small beginning of what should be a bigger effort to add a lot more unit tests. Besides the uninitialized variable `stheta` issue I am very happy with the updates the `kick.f` as well as the new outputted information. I think if this passes it should be merged now and we do a separate PR for the `stheta uninitialized` issue. I ran on quest over a big grid (all bin states) and thing seem to be good.

katiebreivik:
This looks good to me. I'm going to accept/merge the PR and we can move on to the stheta problem.

Issue #114, opened 2019-02-10
totalMass being written multiple times at each step


michaelzevin:
Multiple (identical) values for totalMass are being written to the output file at each step. This doesn't affect the final mass you want at the end, but would be good to fix for clarity when reading the totalMass array.

scottcoughlin2014:
Fixed with #115

PR #115, opened 2019-02-10
Total mass write

katiebreivik:
@michaelzevin found that the total mass was being written to the dat file multiple times. This is a very simple PR that fixes this [issue](https://github.com/COSMIC-PopSynth/COSMIC/issues/114).


katiebreivik:
Oh. It also looks like all the commits from my 'doc_overhaul' branch were copied here as well. This should not add in any problems since the docs have just been updated and the actual cosmic code wasn't really changed.

PR #118, opened 2019-02-13
added in a bpp column for porb

katiebreivik:
This is a very simple PR, just updating bpp array to contain porb column. This will hopefully mitigate any future problems for folks having to remember whether or not bpp has a porb column


scottcoughlin2014:
@katiebreivik totally agree with this column needed to be added to the bpp array. I think this also emphasizes what I talk to @chasebk about and what @michaelzevin originally suggested which is the need for an `updated_the_bpp` separate fortran call so that when we want to add things in the future it does not have to be repeated a billion times like it is right now. Also I had one comment on the PR I think you accidentally added a line you did not mean to.

Issue #120, opened 2019-02-15
Problem with convergence


michaelzevin:
I ran into a problem that I think is regarding the convergence calculations. The failed run is located at 
`/projects/b1011/mzevin/cosmic/r_process/v10_lam0.5`

Below is from the submission log file: 

/home/mjz672/.conda/envs/cosmic-py36/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/home/mjz672/.conda/envs/cosmic-py36/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/home/mjz672/.local/lib/python3.6/site-packages/astropy/stats/histogram.py:209: RuntimeWarning: divide by zero encountered in double_scalars
  Nbins = max(1, np.ceil((dmax - dmin) / dx))
Traceback (most recent call last):
  File "/home/mjz672/.local/lib/python3.6/site-packages/astropy/stats/histogram.py", line 211, in freedman_bin_width
    bins = dmin + dx * np.arange(Nbins + 1)
ValueError: Maximum allowed size exceeded
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/mjz672/.conda/envs/cosmic-py36/bin/runFixedPop", line 293, in <module>
    kstar1_range, kstar2_range, log_file)
  File "/home/mjz672/.conda/envs/cosmic-py36/lib/python3.6/site-packages/cosmic/Match.py", line 163, in perform_convergence
    param_transform(bcm_conv_2[conv_param]).tolist()])
  File "/home/mjz672/.conda/envs/cosmic-py36/lib/python3.6/site-packages/cosmic/Match.py", line 55, in match
    mainHisto, binEdges = astroStats.histogram(np.array(dataCm[0]), bins='knuth')
  File "/home/mjz672/.local/lib/python3.6/site-packages/astropy/stats/histogram.py", line 79, in histogram
    da, bins = knuth_bin_width(a, True)
  File "/home/mjz672/.local/lib/python3.6/site-packages/astropy/stats/histogram.py", line 278, in knuth_bin_width
    dx0, bins0 = freedman_bin_width(data, True)
[...]

PR #122, opened 2019-02-18
Convergence fix

katiebreivik:
Fixing issue: [#120](https://github.com/COSMIC-PopSynth/COSMIC/issues/120) pointed out by @michaelzevin 

This now tries to use knuth binwidths; and if it can't, uses Scott binwidth. It also catches a warning that gets thrown when log10(0) is calculated for porb=0 sources.

### DIFF
--- a/bin/runFixedPop
+++ b/bin/runFixedPop
@@ -28,7 +28,7 @@ import json
 import numpy as np
 import scipy.special as ss
 import pandas as pd
-
+import warnings
 
 from cosmic.sample.initialbinarytable import InitialBinaryTable
 from cosmic import Match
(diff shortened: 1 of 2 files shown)


katiebreivik:
@scottcoughlin2014 I don't want to mess up any of your work with the warnings; should I accept this and https://github.com/COSMIC-PopSynth/COSMIC/pull/123 before or after your PR? They are both small changes.

scottcoughlin2014:
@katiebreivik Go ahead I can merge time into the other branch that is fixing the warning no worries

Scotty

michaelzevin:
@katiebreivik I think there is still an issue with the convergence test for these low-kick runs. I fetched your branch and re-ran some stuff with this commit, and still ran into some problems. You can find the failed error files at /projects/b1011/mzevin/cosmic/r_process/v66.25_lamPols/runfixedpop.param.o* and /projects/b1011/mzevin/cosmic/r_process/v33.125_lamPols/runfixedpop.param.o*

katiebreivik:
Ah- it looks like the binwidths were getting progressively tinier and tinier such that it wigs out the astropy histogram method. Based on my tests a couple of months ago, I had suspected that the binwidths would sort themselves out, but this is clearly not the case here. I will add in a minimum binwidth to the calculation.

PR #124, opened 2019-02-21
Gwinteract compliant changes

scottcoughlin2014:

PR #125, opened 2019-02-21
Convergence fix

katiebreivik:
Ok- I had to reopen this bad boy because @michaelzevin pointed out an issue: https://github.com/COSMIC-PopSynth/COSMIC/issues/120 that needed to be fixed. This was caused by the binwidths getting to be tiny (1-e7) which is not good. Now the binwidth for the convergence histogram has a minimum of 1e-4.

I also removed the merger_type filter since it is now redundant with the bin_state filter.


katiebreivik:
I'm going to quickly merge this in before: https://github.com/COSMIC-PopSynth/COSMIC/pull/124

PR #129, opened 2019-03-06
Fixed polar kick prescription in kick.f

michaelzevin:
Fixed how polar kicks are implemented in kick.f. Now also tracks the total angular change in the orbital plane, which is stores in bkick(20). This meant keeping track of the azimuthal angle of the first SN (omega), and doing some Euler rotations once the second mu/omega are calculated. 

Everything was also coded in Python/Mathematica to double-check, so things should be good to go. 

Added ValueError in runFixedPop when a polar_kick_angle less than 0 or greater than 90 are input.


scottcoughlin2014:
@michaelzevin I assigned myself to this and will get the unit tests in order.

PR #134, opened 2019-03-13
Update docs (And remove all whitespace from source files) No Functional Changes

scottcoughlin2014:


katiebreivik:
@scottcoughlin2014 This looks like you are still in progress; is that correct? If not, let me know and I"ll go through it.

PR #136, opened 2019-03-20
Metallicity in bcm

katiebreivik:
There is something wacky happening here with the evolve test. When I run the test manually in ipython on my laptop everything works. I've verified that the two dataframes are equal several ways too, so it's not the just pandas call that is not working.

I'm not really sure why Travis isn't working, but I know that the tests should all pass.


katiebreivik:
Some more info: since the unit test data had to be update, it is running into the pickling protocol problem again for 2.7.

It also looks like 3.4 is on it's way out, so we should probably just drop it?

@scottcoughlin2014 thoughts?

katiebreivik:
After looking at this more closely, it makes more sense to just alter runFixedPop, so I'm rescinding this PR....

PR #143, opened 2019-03-25
Addition of new CE-related flags

michaelzevin:
This pull request introduces 4 new changes for manipulating CE evolution: 

(*) allows for Qcrit array to be input to specify fixed q_crit values for each kstar

(*) flag for deciding how to implement kicks that are called in the CE (0: old way of doing it, 1: using the mass and SMA prior to CE, 2: using mass and SMA after CE and envelope ejection)

(*) flag for deciding whether stars without a core-envelop boundary merge automatically if they enter a CE

(*) flag for implementing the fitting formulae from Tauris+2015 when evolving helium stars + compact objects in the common envelope


scottcoughlin2014:
@michaelzevin I will get the unit tests updated appropriately.

katiebreivik:
This looks awesome. It looks like the test_single_evolve test needs some work; I expect it's struggling since the initial binary table used for the test doesn't have the necessary CE flags already saved. 

I have run into trouble trying to update the data since apparently there is a difference in the secondary mass produced for some binaries when you use mac vs linux. Maybe this is a random number seed thing?

The other thing that would be awesome to have before we accept this PR is some documentation explaining what to use for each of the flags (and the same thing for the natal kick).

michaelzevin:
Right, we want to sit down one of these days and update all the docs and exemplary Params.ini file.

PR #147, opened 2019-03-27
Bpp array writer

scottcoughlin2014:
@michaelzevin @chase-kimball @katiebreivik I had some time this morning and I just went for it. This has literally crushed my soul. The shear number of random terribly organized calls to the bpp array is staggering. I am not saying this is a permanent solution, but look over it and it is perhaps a good first step. @chase-kimball I know you also where taking a stab at this and maybe we can compare notes and see how we can clean some things up.


chase-kimball:
Hi @scottcoughlin2014,

Thanks for doing this! I'm still in Florida so I haven't had much time to work on this. I got as far as writing a working write_bpp.f but didn't implement it in evolv2. I notice in evolv2 there's a few places where you don't actually pass the mass_pass and kstar_pass variables that you just assigned (see e.g. lines 2097 and 2397). To avoid this sort of thing, and to avoid (further) cluttering up evolv2, I think passing masses and ktypes individually is cleaner than passing the arrays. That's not a hill I'm trying to die on, though. Let me know what you think.

scottcoughlin2014:
> Hi @scottcoughlin2014,
> 
> Thanks for doing this! I'm still in Florida so I haven't had much time to work on this. I got as far as writing a working write_bpp.f but didn't implement it in evolv2. I notice in evolv2 there's a few places where you don't actually pass the mass_pass and kstar_pass variables that you just assigned (see e.g. lines 2097 and 2397). To avoid this sort of thing, and to avoid (further) cluttering up evolv2, I think passing masses and ktypes individually is cleaner than passing the arrays. That's not a hill I'm trying to die on, though. Let me know what you think.

@chase-kimball I agree I think that instead of creating mass_pass and kstar_pass variables that passing m1, m2 and kstar1, kstar2 separately is better. Will work on this

Scotty

scottcoughlin2014:
I have now made sure to pass `mass_pass` and `kstar_pass` are actually used, I am just doing this so that this PR does actually work and then we can consider how best to pass the individual `m1, `m2` instead of arrays.

Issue #130, opened 2019-03-09
Problem when running runfixedpop with different qcrit values


michaelzevin:
I've been doing some runs where I've changed the critical mass ratio at which unstable mass transfer is initiated for naked and evolved naked helium stars (kstar=7,8,9). The runs have been failing part of the way through with the following error: 

Traceback (most recent call last):
  File "/home/mjz672/.conda/envs/cosmic-qcrit/bin/runFixedPop", line 258, in <module>
    bpp, bcm, initCond = Evolve.evolve(initialbinarytable=IBT, BSEDict=BSEDict, nproc=args.nproc, idx=idx, dtp=dtp)
  File "/home/mjz672/.conda/envs/cosmic-qcrit/lib/python3.6/site-packages/cosmic/evolve.py", line 205, in evolve
    bpp_arrays = np.vstack(output[:, 1])
IndexError: too many indices for array
Closing remaining open files:dat_DeltaBurst_13_13.h5...done

The run got through ~11000 binaries before this happened. You can find the working directory at /projects/b1011/mzevin/cosmic/CE_tests/v265_lamPols_qcrit/

michaelzevin:
FYI, this seemed to happen in another run where I changed qcrit as well. The working directory for that run is: 
/projects/b1011/mzevin/cosmic/CE_tests/v265_lamPols_qcrit_endCE

katiebreivik:
I have run into this before as well. I'm not sure exactly what happens, but I suspect that the evolv2 subroutine exits before anything is recorded? It might be worth digging in to see where the goto statements are that send you to the end of the evolution since they could give us a clue as to what is happening to cause the binary evolution to freak out.

We should also probably add in an error catcher that notes the bin_num/IBT for the binary that causes this.

katiebreivik:
I'm running a couple of runs right now to try to reproduce this problem and figure out why BSE is borking...

katiebreivik:
Ok; so I've found a case where this happens:

single_binary = InitialBinaryTable.SingleBinary(m1=0.8810539028092682, m2=0.24953201992853882, porb=1.8193761088893858, ecc=0.4731890565069729, tphysf=9956.942879351827, kstar1=1, kstar2=0, metallicity=0.0001)

[...]

PR #148, opened 2019-03-27
Bpp extend

katiebreivik:
This allows the number of rows in the bpp array to extend out to 1000 per the error (https://github.com/COSMIC-PopSynth/COSMIC/issues/130) pointed out by Mike.

It still has high_res_mw commits in it which I don't understand?


scottcoughlin2014:
@katiebreivik we can talk about rebase later whcih will help with the commit log. I am going to checkout this branch and just do a couple tests. I like it but want to make sure it does not slow things down too much.

Scotty

katiebreivik:
> @katiebreivik we can talk about rebase later whcih will help with the commit log. I am going to checkout this branch and just do a couple tests. I like it but want to make sure it does not slow things down too much.
> 
> Scotty

Sounds good; It shouldn't cause too many problems since it only comes up for a select number of low-probability binaries. (e.g. I had to run ~1e5 double white dwarf binaries to produce the error).

scottcoughlin2014:
Ok so a quick look
original
```
In [2]: %timeit  EvolvedBinaryBPP, EvolvedBinaryBCM, initCond = Evolve.evolve(initialbinarytable=INIT_CONDITIONS, BSEDict=BSEDict)
   ...:                                                                                                                           
701 ms ± 81.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
PR
```
In [2]: %timeit  EvolvedBinaryBPP, EvolvedBinaryBCM, initCond = Evolve.evolve(initialbinarytable=INIT_CONDITIONS, BSEDict=BSEDict)
   ...:                                                                                                                           
929 ms ± 104 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
so an extra 229 ms when running 36 binaries probably no big deal. But just putting it here for posterity.

PR #149, opened 2019-03-28
Convergence typo

katiebreivik:
This PR updates a typo in the convergence calculation. Previously, the bcm data was only transformed using utils.param_transform, when the data should have been transformed with utils.dat_transform.

It also updates runFixedPop to take in match as a command line arg so that you can vary the convergence to fit your needs.


scottcoughlin2014:
@katiebreivik need the FIRE stuff from this PR to be out.

katiebreivik:
All FIRE stuff is out now.

PR #165, opened 2019-04-09
removed flags ceflag and merger

michaelzevin:
removed merger from the ini file and documentation, set it to its default value (-1) in evolv2.f, since it is fed to CMC in CMCpass.


michaelzevin:
We're going to scrap this PR, since we are keeping the CEFLAG as it does surprisingly do something! @scottcoughlin2014 I'll be submitting another PR for this sometime tonight, and I'll close this one at that point.

PR #177, opened 2019-05-10
fixing a common block bug that caused the bhflag to be overwritten

katiebreivik:


katiebreivik:
Ok - So this commit: commit 6d13bcafdd00c7c96e46710e28419d01ea98c80e is the culprit where the common block was reading in Omega... as bhflag. This artificially set all bh kicks to 0. This is illustrated by comparing the current evolve tests to the updated evolution for an example binary:

    bin_num     mass1_binary   mass2_binary     porb           ecc           tphysf
      8644     30.470023     11.401429        6011.99192  0.704389       13700.0


the output from this pr is is:
Screen Shot 2019-05-10 at 2 37 36 PM

and the output from the test data for the develop branch is:
Screen Shot 2019-05-10 at 2 46 27 PM

Note that there is zero kick in the current test data where there should be (see tphys ~ 7.21).

Because of this, I'm updating the test_evol data in the next commit.

katiebreivik:
Ok - so I have no idea how to make this test work but I have verified several times that the test passes on my local machine....

@scottcoughlin2014 do you have any ideas of how to fix this? And make it so we don't have to do this weird runaround every time?

katiebreivik:
This wasn't fully correct and will be overwritten by this PR: #180

PR #184, opened 2019-05-20
Ecsn fix

katiebreivik:
I found that there was a kickscale variable still lurking that just divided sigma for all aic/ecsn kicks by 10 no matter what you did. 

If you specified sigmadiv with a negative number this was overwritten, but the scaling didn't happen. 

I know this PR will break the test_evolve test since the kickscale bug is included in the test data.


katiebreivik:
This hellish PR is actually properly implemented in PR #190

Issue #179, opened 2019-05-14
initial mass sampling is only setup for two final kstar values


katiebreivik:
If you specify more than one kstar value in the intial sampling, you will not get the proper mass filter; e.g. final_kstar1 = [10,11,12,13,14] will filter the masses as kstar_lo=10 and kstar_hi=11. This leads to improper mass distributions which cut off at 20 Msun, so you miss all your BH progenitors.

PR #186, opened 2019-05-21
Init mass samp

katiebreivik:
This PR is needed to fix this issue: https://github.com/COSMIC-PopSynth/COSMIC/issues/179 which cuts off the initial mass sample improperly if more than 2 kstars are provided.


katiebreivik:
This PR is actually properly implemented without the extra commits here: PR #191

PR #192, opened 2019-05-22
Flag checks

michaelzevin:
Added function in utils.py that checks that the values supplied in the BSEDict are viable. This function is called in both evolve and runFixedPop.


katiebreivik:
@michaelzevin I think this looks almost ready to merge; I think we can probably write up some tests pretty quickly that will make the coveralls check also pass. We can talk about it on the call today.

michaelzevin:
As of now, it looks like runFixedPop will already spit out an error if seed_int isn't specified. I'm not sure how this should be handled for the evolve function.

Issue #193, opened 2019-05-28
orbital period intialization is whack


katiebreivik:
Some binaries still start out their evolution when one star is filling its Roche lobe yikes!!

katiebreivik:
Fixed with this pr: https://github.com/COSMIC-PopSynth/COSMIC/pull/197

PR #197, opened 2019-05-29
Fix for issue #193

katiebreivik:
added in a line that takes the maximum and the eccentricity factor needed to compute the Roche Lobe for an eccentric system


katiebreivik:
I'm confirming this because I don't want to disturb @scottcoughlin2014 during thesis weeks. But I also wanted to document that I know that we should normally have a code review!

PR #201, opened 2019-06-04
Multiproc bug

michaelzevin:
This should do the trick, but is bound to break things in the unit tests. We'll probably need to have new unit tests for the multidim sampler and runFixedPop.


michaelzevin:
@katiebreivik this should be good to go

Issue #202, opened 2019-06-04
MCBAGB problem


michaelzevin:
While diving into #195, I found some really wonky stuff is going on with the calculation of the AGB core mass in hrdiag.f. This is causing PISN to _never_ happen, since the core masses that are calculated are tiny. It also seems like for some systems, the mcbagb variable is getting set to the metallicity. 

I ran some pops of BBHs at a couple different metallicities with the PISN flag turned on/off. The initial conditions for these are saved at `/projects/b1095/michaelzevin/cosmic/examples/pisn_systems_init.hdf5` `/projects/b1095/michaelzevin/cosmic/examples/nopisn_systems_init.hdf5`
There's a script that runs the highest mass systems of these initial conditions on Quest here:
`/projects/b1095/michaelzevin/cosmic/examples/pisn_systems.py`, which has a line in it that allows you to chance the metallicity. 

Let's figure this out, since the upper mass gap and PISN are *hot* right now!

michaelzevin:
@katiebreivik @scottcoughlin2014 I cross-checked our hrdiag.f with Carl's from CMC and I can't see anything immediately that jumps out as wrong

michaelzevin:
[hrdiag.txt](https://github.com/COSMIC-PopSynth/COSMIC/files/3287515/hrdiag.txt)

scottcoughlin2014:
With @michaelzevin new branch a `pisn` set to 45. I get this (when sampling initial mass1 between 100 and 200)
 seems good? I will now do it with `pisn` set to 0.
```
(Pdb) bcm.SN_1.value_counts()
0.0    25840
8.0    23760
4.0     1789
9.0      291
Name: SN_1, dtype: int64
```

michaelzevin:
Great! That seems reasonable. @scottcoughlin2014 it would be great to see a plot of the ZAMS mass / remnant mass relation for this run, if that's not too much to ask.

michaelzevin:
Taken care of with #208

Issue #164, opened 2019-04-08
Recording PISN with SN_1, SN_2 (formation1, formation2)


michaelzevin:
So this will actually be a bit trickier than we thought. The problem is that the pisn prescription only gets called in hrdiag, so we will need to change that function to input/output formation1 and formation2, which at this point it doesn't do. 

Do you think this is worth it @katiebreivik @scottcoughlin2014?

michaelzevin:
We decided this would be a good thing to do for a summer student that is also going to implement new PPISN prescriptions, so we'll put it on the back burner for now.

michaelzevin:
Taken care of with #208

PR #208, opened 2019-06-17
PISN bug

michaelzevin:
This PR is regarding #202 #164. First of all, a bug was fixed for naked He stars, where it was still using the mcbagb variable to check whether the system underwent a PISN. Second, we now use SN_1/2=8 and SN_1/2=9 (formation1 and formation2 in the fortran code) to indicate when systems went through PPISN and PISN, respectively. We've also changed the PISN flag to set the maximum core mass (lower-end of the PISN gap) with positive numbers, use the Spera&Mapelli formalism when pisn=-1 (note has typos as of now because Mario miswrote a couple things, the summer student will be fixing it with my/Mario's guidance), and be turned off when pisn=0.


scottcoughlin2014:
This PR fails the unit tests because it fixes a bug which effected the old unit tests. That is the old unit tests involve systems that passed through the faulty line of cold and therefore are nominal badly evolved systems.

PR #210, opened 2019-06-18
Alright trying to use codecov with fortran?

scottcoughlin2014:


scottcoughlin2014:
I am going to merge this so at least the unit test start passing.

PR #211, opened 2019-06-18
Save binary state raction

scottcoughlin2014:


cosmic/utils.py
```diff
@@ -74,28 +74,25 @@ def filter_bpp_bcm(bcm, bpp, method, kstar1_range, kstar2_range):
...
-                bin_num_save.extend(bpp_1.loc[(bpp_1.kstar_1.isin(kstar1_range)) &
-                                              (bpp_1.kstar_2.isin(kstar2_range)) &
-                                              (bpp_1.evol_type == 3)].bin_num.tolist())
-
-            if 2 in use:
-                # SHOULD BE DISRUPTED AND kstar_1.isin(kstar1_range) and kstar_2.isin(kstar2_range)
-                bcm_2 = bcm.loc[bcm.bin_state == 2]
-                bin_num_save.extend(bcm_2.loc[(bcm_2.kstar_1.isin(kstar1_range)) &
-                                              (bcm_2.kstar_2.isin(kstar2_range))].bin_num.tolist())
+
```
katiebreivik: It might be worth adding in a comment here that we need to use the bpp array to select out the systems that actually merge with the kstars that are specified since on first glance it is a bit confusing why you call bcm_0 when it contains binstates 0 and 2. The logic all makes sense and is correct, it just might look a bit weird at first.

scottcoughlin2014: I agree I have tried to address this in my commit

PR #214, opened 2019-06-26
we were cutting off CO WD progenitors

katiebreivik:
I ran into this while getting together the LISA DWD models. CO WD progenitors can come from pretty low initial masses in binaries; especially if there is a stable mass transfer phase.


scottcoughlin2014:
@katiebreivik  looks good but you need to update the 
```
cosmic/tests/test_utils.py::TestUtils::test_min_max_mass_single_kstar FAILED [ 97%]
```
unit test. (its fine that the test evolve fails because of the fixed to PISN)

katiebreivik:
@scottcoughlin2014 This breaks the single evolve tests, but I think a lot of things do... the selection of max/min mass tests now pass

PR #217, opened 2019-07-03
updated max mass for pisn to 135

mkurkowski:


mkurkowski:
Paper from Woosley, S. E. 2017, Astrophys. J., 836, 244
—. 2018, Astrophys. J., 863, 105

PR #223, opened 2019-07-18
constants.py and plots

astroJeff:
Moved constants to constants.py, added plotting.py for plotting evolutionary tracks


astroJeff:
I added to the plotting script the options to plot multiple binaries.

PR #227, opened 2019-07-26
Qcrit binary c

katiebreivik:
Ok! So this a pretty big PR. The main idea here is to add a few different models for the critical mass ratio at which RLO mass transfer remains stable or becomes unstable and creates a common envelope. The 5 models are:
qcflag = 0: default BSE qcrits, not using Hjellming & Webbink 1987 for giants
qcflag = 1: default BSE qcrits, using Hjellming and Webbink 1987 for giants
qcflag = 2: binary_c qcrits following Table 2 in Claeys+2014, not using Hjellming and Webbink 1987 for giants. 
qcflag = 3: binary_c qcrits following Table 2 in Claeys+2014, using Hjellming and Webbink 1987 for giants. 
qcflag = 4: use qcrit_array values. 

Note that qcflag 2/3 use equation 10-11 of Claeys+2014 to compute mass transfer rate factor: f. 

I have updated the error checker to throw errors if the qcflag and qcrit_array are wrong. 

This will definitely break Travis CI....


katiebreivik:
Here are some cute plots that show the impact of the different models. BSE models are qcflag=1, binary_c models are qcflag=3. Both plots are the result of a simulated grid and show either: the initial parameters for binaries that form DWDs or the DWD parameters at formation. The binary_c models give us a factor of 4 more DWDs that have much higher primary masses. Yay science!

[initial_params_for_DWD_binaries.pdf](https://github.com/COSMIC-
PopSynth/COSMIC/files/3436868/initial_params_for_DWD_binaries.pdf)
[DWD_formation_params.pdf](https://github.com/COSMIC-PopSynth/COSMIC/files/3436869/DWD_formation_params.pdf)

Issue #224, opened 2019-07-22
current documentation describing units is not robust/consistent


celiotine:

scottcoughlin2014:
All of these things needs units defined
```
bpp_columns = ['tphys', 'mass_1', 'mass_2', 'kstar_1', 'kstar_2' ,
               'sep', 'porb', 'ecc', 'RROL_1', 'RROL_2', 'evol_type',
               'Vsys_1', 'Vsys_2', 'SNkick', 'SNtheta',
               'aj_1', 'aj_2', 'tms_1', 'tms_2',
               'massc_1', 'massc_2', 'rad_1', 'rad_2',
               'bin_num']
```
```
bcm_columns = ['tphys', 'kstar_1', 'mass0_1', 'mass_1', 'lumin_1', 'rad_1',
               'teff_1', 'massc_1', 'radc_1', 'menv_1', 'renv_1', 'epoch_1',
               'ospin_1', 'deltam_1', 'RROL_1', 'kstar_2', 'mass0_2', 'mass_2',
               'lumin_2', 'rad_2', 'teff_2', 'massc_2', 'radc_2', 'menv_2',
               'renv_2', 'epoch_2', 'ospin_2', 'deltam_2', 'RROL_2',
               'porb', 'sep', 'ecc', 'B_0_1', 'B_0_2',
               'SNkick_1', 'SNkick_2', 'Vsys_final', 'SNtheta_final',
               'SN_1', 'SN_2', 'bin_state', 'merger_type', 'bin_num']
```

michaelzevin:
bpp_columns
- [ ] tphys
- [ ] mass_1
- [ ] mass_2
- [ ] kstar_1
- [ ] kstar_2
- [ ] sep
- [ ] porb
- [ ] ecc
- [ ] RROL_1
- [ ] RROL_2
- [ ] evol_type
- [ ] Vsys_1
- [ ] Vsys_2
- [ ] SNkick
- [ ] SNtheta
- [ ] aj_1
- [ ] aj_2
- [ ] tms_1
- [ ] tms_2
- [ ] massc_1
- [ ] massc_2
- [ ] rad_1
- [ ] rad_2
- [ ] bin_num

michaelzevin:
bcm_columns

- [ ] tphys
- [ ] kstar_1
- [ ] mass0_1
- [ ] mass_1
- [ ] lumin_1
- [ ] rad_1
- [ ] teff_1
- [ ] massc_1
[...]

Issue #172, opened 2019-04-15
Installation should note that if you want to use jupyter you need to install it in your conda env


katiebreivik:
If folks create a new cosmic env and don't also install jupyter, it will not obviously break and will pull from the globally installed package if it exists.

scottcoughlin2014:
I will add a WARNING message in the installation page on the docs in my upcoming PR

michaelzevin:
Addressed in #230

Issue #161, opened 2019-04-08
Docs probably too heavily emphasize Milky Way populations (rephrasing needed)


scottcoughlin2014:
just say binary populations or something

Issue #128, opened 2019-03-01
install problems with conda gcc


katiebreivik:
It looks like the conda gcc doesn't play nicely anymore after a mac upgrade to mojave. I have verified that the install works if you have the developer tools gcc version, and jeff has it working with a homebrew gcc. 

Here is an example of someone else running into this problem: https://github.com/openai/mujoco-py/issues/331

michaelzevin:
Totally fixed!

ReneKat:
@michaelzevin 
I'm having this issue myself and the `conda update gcc` doesn't fix it. Any guidance would be appreciated. Thank you!

I am running OSX Mojave 10.14.6 and have just updated Xcode to 11.3.1 trying to solve this gcc compilation error. 

`pip install git+https://github.com/danielfrg/tsne.git`

```
Collecting git+https://github.com/danielfrg/tsne.git
  Cloning https://github.com/danielfrg/tsne.git to /private/var/folders/5s/r0xdrzyj2dl_43mj6jh0cmzc0000gp/T/pip-req-build-wu_d_54h
  Running command git clone -q https://github.com/danielfrg/tsne.git /private/var/folders/5s/r0xdrzyj2dl_43mj6jh0cmzc0000gp/T/pip-req-build-wu_d_54h
Requirement already satisfied: Cython>=0.19.1 in /Users/Owl/miniconda3/envs/autometa/lib/python3.8/site-packages (from tsne==0.1.8) (0.29.15)
Requirement already satisfied: numpy>=1.7.1 in /Users/Owl/miniconda3/envs/autometa/lib/python3.8/site-packages (from tsne==0.1.8) (1.18.1)
Requirement already satisfied: scipy>=0.12.0 in /Users/Owl/miniconda3/envs/autometa/lib/python3.8/site-packages (from tsne==0.1.8) (1.4.1)
Building wheels for collected packages: tsne
  Building wheel for tsne (setup.py) ... error
  ERROR: Command errored out with exit status 1:
[...]

PR #230, opened 2019-07-31
WIP: Adding bse flag documentation

scottcoughlin2014:
This should start to address the majority of documentation issues outstanding before the next release. I will link this PR to all the ones it is intended to address.


scottcoughlin2014:
#75  #84 #95 #128 #161 #172 #224

PR #235, opened 2019-08-01
units overhaul

katiebreivik:
changing all units to match those in the docs, the main changes are to make all log10 values not log10 and to write the orbital period as days


katiebreivik:
This will absolutely wreck the single evolve

PR #243, opened 2019-08-05
Added the ability to pass a path to the inifile contain the BSE flags…

scottcoughlin2014:
… in order to run evolve. Also I really revamped the logic of the three ways that one can pass the BSE flag information to Evolve.evolve through better documentation and clearer key word argument passing. Moreover in order to stop getting the pandas SettingWithCopyWarning I changed the way we assign the passed flags as columns to the DF. I think it is better and the warning will stop happening.

### DIFF
--- a/cosmic/evolve.py
+++ b/cosmic/evolve.py
@@ -29,20 +29,22 @@ from gwpy.utils import mp as mp_utils
 import numpy as np
 import pandas as pd
 import json
+import warnings
+import os
 
 __author__ = 'Katelyn Breivik <katie.breivik@gmail.com>'
 __credits__ = 'Scott Coughlin <scott.coughlin@ligo.org>'
 __all__ = ['Evolve']
 
 
-bpp_columns = ['tphys', 'mass_1', 'mass_2', 'kstar_1', 'kstar_2' ,
+BPP_COLUMNS = ['tphys', 'mass_1', 'mass_2', 'kstar_1', 'kstar_2' ,
                'sep', 'porb', 'ecc', 'RROL_1', 'RROL_2', 'evol_type',
                'Vsys_1', 'Vsys_2', 'SNkick', 'SNtheta',
                'aj_1', 'aj_2', 'tms_1', 'tms_2',
                'massc_1', 'massc_2', 'rad_1', 'rad_2',
                'bin_num']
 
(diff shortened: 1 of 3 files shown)


michaelzevin:
@scottcoughlin2014 do you know why the unit test failed for this? I can't see why they would.

scottcoughlin2014:
> Looks pretty good to me @scottcoughlin2014. Before approving, did you check that this works when you provide a Params.ini file? You should check this when (a) there are already flags set in the initial binary table and (b) when there are not flags set in the initial binary table.
> 
> Also, do the flags in the initialbinarytable always take precedence whether you specify a BSEDict or Params.ini?

Hey Mike,

In the two unit tests I added to `test_evolve` I have included all three scenarios including passing the `Params.ini` file (in fact the one that is in the `examples` folder).

Also in the doc strings I have added the answer
```
        NUMBER 2: PASS A PANDAS DATA FRAME WITH PARAMS DEFINED AS COLUMNS
             All you need is the initialbinarytable with columns,
             If you pass both a dictionary of flags and/or a inifile
             and a table with the columns, the column values for the flags
             take precedent
```

scottcoughlin2014:
> @scottcoughlin2014 do you know why the unit test failed for this? I can't see why they would.

@michaelzevin Thisis related to the `porb` unit changes (see slack discussion) Once Katie does another PR to address it the unit tests should not fail (i.e. it fails for an unrelated reason)

michaelzevin:
@scottcoughlin2014 per our discussion in the COSMIC meeting today, we are going to make 2 changes: 

1) if BSEDict or Params.ini are specified in the kwargs, they will take precedence over what is specified in the initialbinarytable. 
2) Add a warning if BSEDict or Params.ini are trying to overwrite a flag that is already set in initialbinarytable.

scottcoughlin2014:
> @scottcoughlin2014 per our discussion in the COSMIC meeting today, we are going to make 2 changes:
> 
>     1. if BSEDict or Params.ini are specified in the kwargs, they will take precedence over what is specified in the initialbinarytable.
> 
>     2. Add a warning if BSEDict or Params.ini are trying to overwrite a flag that is already set in initialbinarytable.

@michaelzevin I have switched the logic of the ```kwargs``` and immensely cleaned up the logic of the code, far fewer lines and now when you want to add a flag to pass to BSE I think it will be easier.

PR #245, opened 2019-08-06
This PR will ensure that the natal kick is written appropriately ever…

scottcoughlin2014:
…y time there is a SN during the evoltuion. It explicitly decribes all possible scenarios now. This ensures that things will work as expected


scottcoughlin2014:
@michaelzevin has approved. Going to merge

PR #251, opened 2019-08-07
Bin frac

katiebreivik:
Ok - now we track the mass and number of binaries and single stars correctly. They are also now all stored in the dataframe.


bin/runFixedPop
```diff
@@ -216,18 +216,24 @@ if __name__ == '__main__':
...
-            IBT, sampled_mass, n_samp = InitialBinaryTable.sampler(args.initial_samp, kstar1_range, kstar2_range, rand_seed, args.nproc, SFH_model, component_age, args.metallicity, args.Nstep)
+            IBT, mass_singles, mass_binaries, n_singles, n_binaries= InitialBinaryTable.sampler(args.initial_samp, kstar1_range, kstar2_range, rand_seed, args.nproc, SFH_model, component_age, args.metallicity, args.Nstep)
 
         # Log the total sampled mass from the initial binary sample
         # for future Galactic occurence rate calculation
-        total_sampled_mass += sampled_mass
+        total_mass_singles += mass_singles
+        total_mass_binaries += mass_binaries
+        total_n_singles += n_singles
+        total_n_binaries += n_binaries
```
michaelzevin: I don't think these lines will work, since you never initialized the variables `total_mass_singles`, `total_mass_binaries`, `total_n_singles`, and `total_n_binaries`. Shouldn't you do this at the try/except statement at 169?

cosmic/sample/sampler/independent.py
```diff
@@ -226,17 +265,21 @@ def binary_select(self, primary_mass, model='half'):
...
+                binary_fraction = 1/2.0 + 1/4.0 * np.log10(primary_mass)
+                binary_choose =  np.random.uniform(0, 1.0, primary_mass.size)
 
-            binaryIdx, = np.where(binary_fraction > binary_choose)
-            singleIdx, = np.where(binary_fraction < binary_choose)
+                binaryIdx, = np.where(binary_fraction > binary_choose)
+                singleIdx, = np.where(binary_fraction < binary_choose)
+            else:
+                raise Error('You have supplied a non-supported binary fraction model. Please choose vanHaaften or a float')
+        elif type(model) == float:
```
michaelzevin: In this line, you should force the binary fraction to be 0 <= bin_frac <= 1, otherwise raise a value error

bin/runFixedPop
```diff
@@ -216,18 +216,24 @@ if __name__ == '__main__':
...
 
         # save the total_sampled_mass so far
-        dat_store.append('totalMass', pd.DataFrame([total_sampled_mass]))
-        log_file.write("The total mass sampled so far is: {0}\n".format(total_sampled_mass))
+        dat_store.append('mass_singles', pd.DataFrame([total_mass_singles]))
+        dat_store.append('mass_binaries', pd.DataFrame([total_mass_binaries]))
+        dat_store.append('n_singles', pd.DataFrame([total_n_singles]))
+        dat_store.append('n_binaries', pd.DataFrame([total_n_binaries]))
```
michaelzevin: Through redundant, I think it would be useful to store the total mass and total number of stars (n_singles + 2*n_binaries) as well.

katiebreivik:
Thanks mike- I will do this tomorrow morning. 

——————————
Katelyn Breivik
CITA Fellow
katiebreivik.github.io
——————————

Issue #253, opened 2019-08-08
PISN SN type getting set incorrectly


michaelzevin:
For the run at /projects/b1095/michaelzevin/sgrb/pop_models/Z10_Tauris1/, it seems like many of the DNS binaries have SN_2=8, which is the code for pulsational pair instability. Since these are NS progenitors, none of these should have gone through this.

scottcoughlin2014:
@michaelzevin LOL turns out formation=8 was already in BSE.....................
```
 cosmic/src/comenv.f 

574                if(KW1.eq.13.and.KW2.ge.13.and.ussn.eq.1)then
575                   if(KW1i.ge.7.and.KW1i.le.9)then
576                      if(sigma.gt.0.d0.and.sigmadiv.gt.0.d0)then
577                         sigma = sigmahold/sigmadiv
578                         sigma = -sigma
579                      else
580                         sigma = -1.d0*sigmadiv
581                      endif
582                   formation1 = 8
583                   endif
584                endif

```

scottcoughlin2014:
#255 fixes this

PR #255, opened 2019-08-08
Changing pisn 253

scottcoughlin2014:
Addresses #253 redoes relation of the integer value of `formation` and the type of SN it relates to


scottcoughlin2014:
> This looks good, but I am perplexed how you updated the docs to match this convention and it doesn't show up in the commits?

@katiebreivik lol I did this on the side. but I will merge this here so that they are in sync.

katiebreivik:
Thanks!

michaelzevin:
@scottcoughlin2014 Just ran a new DNS population and the new numbers are looking good.

PR #256, opened 2019-08-08
Metallicity dependence on wind mass-loss

michaelzevin:
I've finished what I can do with the metallicity-dependent mass loss. This PR adds a new flag (called eddlimflag) that applies a special metallicity dependence to wind mass loss to stars near the Eddington limit. It seems to be producing almost what is in Giacobbo et al. 2018, with the exception of a small dip in remnant masses near about 60 Msun, which from what I can tell are due to thermally pulsating AGB stars. 

After hours of digging into this and comparing with MOBSE, I have to call it quits on trying to figure out this discrepancy. I'll attach an image of the Mzams/Mrem plot for clarification, but I'm ready to call this PR good to go and move on.


michaelzevin:
MzamsMrem_newflag

scottcoughlin2014:
@michaelzevin good work sticking this out. I will be happy to review.

PR #258, opened 2019-08-09
fixed cehestarflag=2 by ensuring the orbital period is in the range o…

michaelzevin:
…f validity of the fitting formulae and that the derived post-CE mass is greater than the mass of the stellar core. I would still recommend using instead cehestarflag=1, though, since this is not self consistent.

### DIFF
--- a/cosmic/src/comenv.f
+++ b/cosmic/src/comenv.f
@@ -458,12 +458,16 @@
      &                  SEP**(3d0/2d0)
                     Mcf = (1d0/(400d0*Porbi)+0.49d0)*MF -
      &                  ((0.016d0/Porbi) - 0.106d0)
-                    if(Porbi.le.2d0)then
+                    if(Porbi.le.2d0.and.Porbi.gt.0.06d0)then
                         Menvf = 0.18d0*Porbi**(0.45d0) *
      &                      (LOG(Mcf**4d0) - 1.05d0)
-                    else
+                    elseif(Porbi.gt.2.d0)then
                         Menvf = Mcf*(LOG(Porbi**(-0.2d0))+1) +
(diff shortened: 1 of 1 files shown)


katiebreivik:
@michaelzevin did you add a note in the docs where we define the cehestarflag? If not and you need this done asap, I can do it, but I think we should try to keep the docs up to date since they are so nice :-)

michaelzevin:
@katiebreivik @scottcoughlin2014 how to I change to docs again?

scottcoughlin2014:
@michaelzevin if this is just updating the information here: https://cosmic-popsynth.github.io/docs/latest/inifile/index.html for the section `cehestarflag` then just put a blurb here explaining thechange to `cehestarflag=2` and I will add it to the docs and then push it online and we can merge this.

scottcoughlin2014:
@michaelzevin  Once we merge the update to the documentation into `develop` then in general you can go into

```
vi docs/inifile/index.rst 
```
and update the relevant flags info.

michaelzevin:
Ok @scottcoughlin2014 I might be dumb dumb but I can't figure out how to edit that page. Here's what you should put in for the blurb: 

; cehestarflag uses fitting formulae from Tauris, Langer, and Podsiadlowski, 2015, MNRAS, 451 for evolving RLO systems with a helium star donor and compact object accretor
; this flag will override choice made by cekickflag if set
; if cehestarflag=2, the range of validity is Porb,i > 0.06 days. The final mass fits are also limited to be greater than the core mass of the star.

michaelzevin:
Also a note @scottcoughlin2014: as they currently stand, the docs are not account for linebreaks.

katiebreivik:
I can update with @michaelzevin's blurb now.

Issue #248, opened 2019-08-07
np.isclose is not working to ignore columns for convergence that do not matter


scottcoughlin2014:
things like `ecc` and `sep` for merged systems are still being tracked.

katiebreivik:
https://github.com/COSMIC-PopSynth/COSMIC/pull/260 fixes this

PR #260, opened 2019-08-11
Convergence ecc

katiebreivik:
This is a hefty PR that adds several convergence flags and fixes the bugs in: https://github.com/COSMIC-PopSynth/COSMIC/issues/248

I've held off on the filtering flags since I want to chat with folks at the call on Tuesday about what to do there-


katiebreivik:
Alright @scottcoughlin2014 I think this is ready for review.

Issue #196, opened 2019-05-29
tidal strength scaling factors


ageller:
A few studies (including mine and StarTrack) have shown that the normal Hut (+BSE method) convective tides are too weak to explain observed MS-MS binary populations, at least for solar-type stars.  I’ve fudged this in my BSE code by including a scaling factor in front of the tcqr equation in evolv2.f .  For instance, in your master branch, you could include a factor for the convective damping on line 908 of evolv2.f (tcqr = fprimc * 2.d0*f*q …, where fprimc is the input scaling factor).  Likewise, for completeness, you could include factors for the radiative (line 898) and degenerate (line 915) tides, though I don’t know if observations suggest you need such factors there.  There are also additional similar tcqr equations further down in evolve2.f, around line 2800 that should also have these same scaling factors applied.  

These should be available on input from python.  Maybe you would even want this to be a function of the stellar type.

scottcoughlin2014:
@ageller we are trying to incorporate this request before our release of the code. Wea re wrondering if just having a scaling factor is the way to go or a scaling factor array that is kstar dependent (so a like 16 vector array like we do for what q_crit value to use).

ageller:
A kstar dependent value (input as an array) would be ideal.  Thanks.

katiebreivik:
https://github.com/COSMIC-PopSynth/COSMIC/pull/263 fixes this

PR #263, opened 2019-08-14
Tidal strength scaling factors 196

scottcoughlin2014:
#196 This addresses Aarons request and fixes a bug in the convective tides that aaron also pointed out.

`fprimc_array` should be added everywhere that is needed. Once it is approved I will make sure to updated the `latest` docs with the new flag in the doc example and inifile info.


cosmic/tests/data/Params.ini
```diff
@@ -279,3 +279,7 @@ bconst=-3000
 ; default=-1000
 ck=-1000
 
+; fprimc controls the scaling factor for convective tides
+; each item is set individually for its associated kstar
+; default is the original BSE value of 2.0 for every kstar
```
katiebreivik: If the default is that everything is 2.0, why is fprimc_array set to all 1.0?

docs/inifile/index.rst
```diff
@@ -554,20 +554,26 @@ Each of the below sections lists the valid options and a description of what the
...
+                  **default=0.001**
+``bconst``        *bconst* related to magnetic field evolution of pulsars. Implemented by Paul Kiel -- see Section 3 of `Kiel et al. 2008 <https://academic.oup.com/mnras/article/388/1/393/1013977>`_.
+                  **default=-3000**
+``ck``            *ck* related to magnetic field evolution of pulsars, . Implemented by Paul Kiel -- see Section 3 of `Kiel et al. 2008 <https://academic.oup.com/mnras/article/388/1/393/1013977>`_.
+                  **default=-1000**
+
+``fprimc_array``  fprimc_array controls the scaling factor for convective tides
+                  each item is set individually for its associated kstar
+                  default is the original BSE value of 2.0 for every kstar
+
```
katiebreivik: update here as well if we need to update the default.

docs/inifile/index.rst
```diff
@@ -598,3 +604,8 @@ Each of the below sections lists the valid options and a description of what the
     ; ck is related to magnetic field evolution of pulsars, see Kiel+2008
     ; default=-1000
     ck=-1000
+
+    ; fprimc controls the scaling factor for convective tides
+    ; each item is set individually for its associated kstar
+    ; default is the original BSE value of 2.0 for every kstar
```
katiebreivik: update here as well if we need to update the default

examples/Params.ini
```diff
@@ -279,3 +279,7 @@ bconst=-3000
 ; default=-1000
 ck=-1000
 
+; fprimc controls the scaling factor for convective tides
+; each item is set individually for its associated kstar
+; default is the original BSE value of 2.0 for every kstar
```
katiebreivik: update here if default changes

PR #264, opened 2019-08-15
System binfrac

michaelzevin:
Hey team, I think I've got the binary fraction tracking good to go. Did tests with independent and multidim samplers and it seems to be working fine. The system-specific binary fraction probability is now output in the initCond file.


katiebreivik:
Hey @michaelzevin - It looks like this needs to have the initial_conditions_for_testing.hdf5 file updated based on the PR @scottcoughlin2014 just merged. https://github.com/COSMIC-PopSynth/COSMIC/pull/263

I think the best option is for you to fetch from the current develop branch that contains the last PR and re-fix the test file, but @scottcoughlin2014 might have better thoughts here...

scottcoughlin2014:
Assuming my review goes fine in terms of the updated code, then I can just do this change of the file manually and merge at same time.

Issue #270, opened 2019-08-20
seg fault is happening. yikes!!


katiebreivik:
We have isolated a 100 binary inifile that is causing a seg fault. We are fixing it on a cosmic call on 20 aug 19

katiebreivik:
Ok! so it looks like this was caused because we weren't sending in ecsn to mix the two times that it is called in evolv2.f

Maybe there are other places that there could be a seg fault?

PR #271, opened 2019-08-20
ecsn was not being passsed to mix

scottcoughlin2014:


scottcoughlin2014:
#270 this will fix the segfault for this situation there may be others

PR #275, opened 2019-08-25
Removing matplotlib and gwpy dependecies as they are not necessary no…

scottcoughlin2014:
…w that I use the gwpy file directly (keeping the copyright info to be Duncan's of course) and matplotlib was never really used. Anyways this should make installation and putting it on conda a little more straightforward


scottcoughlin2014:
@katiebreivik I am going to merge this since there is no functional change I am just changing the dependencies. But if you want to see where I moved the multiprocessing stuff from gwpy you can see it there.

PR #286, opened 2019-09-17
changed up common variables in evolve.py, including how they are set.…

scottcoughlin2014:
… common variables are no longer passed to evolv2


scottcoughlin2014:
I think this should be run under gdb to make sure that everything is good in the fortran in terms of everything is set right. That, or if someone has a big intiial conditions table they already ran and now the BPP outputs for and could run again with this branch that would be good. @katiebreivik

katiebreivik:
Thanks @scottcoughlin2014 I will take a look at this tomorrow with an initC file that I had previously run to check if the bpp arrays are the same. Will check back in once that's done.

cosmic/src/comenv.f
```diff
@@ -87,24 +86,24 @@ SUBROUTINE COMENV(M01,M1,MC1,AJ1,JSPIN1,KW1,
...
       RZAMS = RZAMSF(M01)
 *
 * Decide which CE prescription to use based on LAMBDA flag
 * MJZ: NOTE - Nanjing lambda prescription DOES NOT WORK!
 *
-      IF(LAMBDA.EQ.1.0)THEN
-         LAMB1 = CELAMF(KW,M01,L1,R1,RZAMS,MENVD,LAMBDA)
-      ELSEIF(LAMBDA.LT.0.d0)THEN
-         LAMB1 = -1.0*LAMBDA
+      IF(LAMBDAF.EQ.1.0)THEN
```
katiebreivik: @carlrodriguez should make a not that we have changed LAMBDA to LAMBDAF

carlrodriguez: Good to know, thanks.  I'm going to go through the const_bse.h file and every common variable to make sure there's still 1-to-1 correspondance.

scottcoughlin2014:
> Ok; so after checking this on a population of 5281 binaries, there are 9 systems which don't match exactly. However, the differences are all in the thousandths place or greater for the masses, separations, eccentricities, orbital periods, and roche radii so I think we are safe to move forward.

I also checked in gdb and things look good. I will mergfe then!

PR #290, opened 2019-09-19
Docs plotting and final cleanup

katiebreivik:
Alright! Now we have docs for the plots and they are in a state that I'm happy with.


katiebreivik:
@scottcoughlin2014 The only thing that is left for this which I am totally baffled by is that the miscellaneous flags section table is overflowing but I can't find anything that suggests that it should?

Issue #297, opened 2019-09-25
0.1 Factor of Eq 80 of Hurley needs to be a flag


scottcoughlin2014:
Mario has called out equation 80 of Hurley. The factor of 0.1 is not going to be a good factor when we integrate with CMC. We instead should make this a flag (yay flags) that can be set by the user.

FOR EXAMPLE THIS IS BAD
```
In [84]: In [3]: single_binary = InitialBinaryTable.InitialBinaries(m1=140.0, m2=0.1, porb=4.795757, ecc=0.03, tphysf=13700.0, kstar1=1,
    ...:  kstar2=1, metallicity=0.002)                                                                                                  

In [85]: In [6]: bpp, bcm, initC = Evolve.evolve(initialbinarytable=single_binary, BSEDict=BSEDict, dtp=0.01)                           

In [86]: bpp                                                                                                                            
Out[86]: 
        tphys      mass_1  mass_2  kstar_1  kstar_2       sep  ...         tms_2    massc_1  massc_2        rad_1     rad_2  bin_num
0    2.762072  123.953937     0.1      1.0      0.0  42.99578  ...  2.662638e+06   0.000000      0.0    33.941325  0.141798        0
0    2.762072  124.053937     0.1      1.0     15.0   0.00000  ...  2.662638e+06   0.000000      0.0    33.941325  0.141798        0
0    3.346692  122.156696     0.0      1.0     15.0   0.00000  ...  2.662638e+06   0.000000      0.0    13.610769  0.141798        0
0    5.855395   95.698415     0.0      2.0     15.0   0.00000  ...  2.662638e+06  44.565727      0.0    42.546671  0.141798        0
0    5.857622   95.385637     0.0      4.0     15.0   0.00000  ...  2.662638e+06  45.456823      0.0  7136.456170  0.141798        0
0    6.176004   47.628410     0.0      7.0     15.0   0.00000  ...  2.662638e+06   0.000000      0.0     2.444429  0.141798        0
0    6.198909   46.406641     0.0      7.0     15.0   0.00000  ...  2.662638e+06   0.000000      0.0     2.382399  0.141798        0
0    6.198909   40.500000     0.0     14.0     15.0   0.00000  ...  2.662638e+06  40.500000      0.0     0.000172  0.141798        0
0  206.180000   40.500000     0.0     14.0     15.0   0.00000  ...  2.662638e+06  40.500000      0.0     0.000172  0.141798        0

[9 rows x 24 columns]
[...]

PR #299, opened 2019-09-25
Rejuvenation flag

katiebreivik:
This addresses Issue 297: https://github.com/COSMIC-PopSynth/COSMIC/issues/297


katiebreivik:
😢 The test_single_evolve_table test failed but I don't understand why because it didn't fail on my own laptop.

PR #304, opened 2019-09-27
added _evolve_multi_system

mcdigman:
I added a function _evolve_multi_system which allows binaries to be evolved in batches of size n_per_block, a new optional kwarg for Evolve. Using this function removes the limit on Nstep of approximately 30000 set by the maximum size of the Multiprocessing queue. For example, with n_per_block=200, the limit on the number of systems per step becomes 6000000. The output processing which occurs every Nstep systems is extremely wall-time intensive on a supercomputer, partly because it takes place on only a single core, such that the limitation of ~30000 systems is a major performance bottleneck when running a large number of systems. This function significantly improves the practicability of evolving 100s of millions of systems on a supercomputer. 

I have also added appropriate test cases to test_evolve.py to verify the output while using the function is identical.

### DIFF
--- a/cosmic/evolve.py
+++ b/cosmic/evolve.py
@@ -155,6 +155,11 @@ class Evolve(object):
             timestep size in Myr for bcm output where tphysf
             is total evolution time in Myr
 
+        n_per_block : `int`, optional, default: -1
+            number of systems to evolve in a block with 
+            _evolve_multi_system, to allow larger multiprocessing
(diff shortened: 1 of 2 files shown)


scottcoughlin2014:
Thanks @mcdigman I will review this shortly and I think we will be able to see this feature in `3.0.0` release we are trying to officially deploy this weekend

PR #311, opened 2019-10-01
Fixed lambdaf issue

carlrodriguez:
Reverted COSMIC to use lambdaf=0.5 instead of 1.0 and changed the defaults as well.  This fixes the NaNs we were seeing in some binary evolution


katiebreivik:
Looks good. Thanks @carlrodriguez 🥇

PR #315, opened 2019-10-02
Evolve hard code flags

katiebreivik:
Ok; removed the hard coded stuff in evolv2.f. Everything left now is just pulsar stuff that doesn't change so there is no use for different flags.

Also, the tests might break because I changed ST_tide back to 1. Not really sure why I had it at 0? Must have been when comparing to SeBa a while back.


katiebreivik:
Man this was a journey!

katiebreivik:
Ran a successful cosmic-pop call for 1e5 binaries with kstars 0-14. I think this is good to go.

PR #317, opened 2019-10-03
Fixed Common block nightmare

carlrodriguez:
Made the common block do what it's supposed to


katiebreivik:
Yay! Love to have flags that are set to what we want them to be set to!

PR #342, opened 2020-01-15
Allowing the ability to restart a bianry mid evolution

scottcoughlin2014:
Hey All,

The unit tests for evolving the binary now fail but I know how to fix it. I think it would be good to discuss this PR so we can get it merged for Camille to start using. There are definitely parts of this functionality I would like to see improved but I think that can occur after this more straightforward PR is approved. @katiebreivik @astroJeff @michaelzevin. I think in addition to unit testing this new feature (and adding documentation) it would be great if people could test it out on some evolution they have from using COSMIC 3.2. An example of how I first do an evolution and then test doing that same evolution but at different starting points is below.

```
from cosmic.sample.initialbinarytable import InitialBinaryTable

from cosmic.evolve import Evolve

import pandas

single_binary = InitialBinaryTable.InitialBinaries(m1=85.543645, m2=84.99784, porb=446.795757, ecc=0.448872, tphysf=13700.0, kstar1=1, kstar2=1, metallicity=0.002)

BSEDict = {'xi': 1.0, 'bhflag': 1, 'neta': 0.5, 'windflag': 3, 'wdflag': 1, 'alpha1': 1.0, 'pts1': 0.001, 'pts3': 0.02, 'pts2': 0.01, 'epsnov': 0.001, 'hewind': 0.5, 'ck': -1000, 'bwind': 0.0, 'lambdaf': 0.5, 'mxns': 2.5, 'beta': 0.125, 'tflag': 1, 'acc2': 1.5, 'nsflag': 3, 'ceflag': 0, 'eddfac': 1.0, 'ifflag': 0, 'bconst': -3000, 'sigma': 265.0, 'gamma': -1.0, 'pisn': 45.0, 'natal_kick_array' : [-100.0,-100.0,-100.0,-100.0,-100.0,-100.0], 'bhsigmafrac' : 1.0, 'polar_kick_angle' : 90, 'qcrit_array' : [0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0], 'cekickflag' : 2, 'cehestarflag' : 0, 'cemergeflag' : 0, 'ecsn' : 2.5, 'ecsn_mlow' : 1.4, 'aic' : 1, 'ussn' : 0, 'sigmadiv' :-20.0, 'qcflag' : 2, 'eddlimflag' : 0, 'fprimc_array' : [2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0,2.0/21.0], 'bhspinflag' : 0, 'bhspinmag' : 0.0, 'rejuv_fac' : 1.0, 'rejuvflag' : 0, 'htpmb' : 1, 'ST_cr' : 1, 'ST_tide' : 0, 'bdecayfac' : 1, 'randomseed' : -1235453}

for i in [3, 7, 11]:
    bpp, bcm, initC = Evolve.evolve(initialbinarytable=single_binary, BSEDict=BSEDict)

    for column in bpp.columns:
        initC = initC.assign(**{column:bpp.iloc[i][column]})

    bpp_mid, bcm_mid, initC_mid = Evolve.evolve(initialbinarytable=initC)

    print("From beginning")
    print(bpp[['tphys','mass1_binary','mass2_binary','kstar_1','kstar_2','evol_type', ]])
    print(bcm[['Vsys_final','SNkick_1', 'SNkick_2']])


    print("Started in middle at Index {0}".format(i))
    print(bpp_mid[['tphys','mass1_binary','mass2_binary','kstar_1','kstar_2','evol_type',]])
    print(bcm_mid[['Vsys_final','SNkick_1', 'SNkick_2']])
```


cosmic/src/evolv2.f
```diff
@@ -465,11 +464,7 @@ SUBROUTINE evolv2(kstar,mass,tb,ecc,z,tphysf,
 *
 * On the first entry the previous timestep is zero to prevent mass loss.
 *
-      if(using_cmc.eq.1)then
-          dtm = 0.d0
-      else
-          dtm = 1.d0
-      endif
+      dtm = 0.d0
       delet = 0.d0
```
scottcoughlin2014: @katiebreivik @astroJeff  this is the main thing I need to make sure is OK to do.

katiebreivik:
So I'm seeing a weird error here:
If I run the standard example from the docs, I get the right output (within a few decimal points which is good). However, if I restart at a point in the evolution where the kstar type isn't 1, it is forced to one:

Screen Shot 2020-01-16 at 5 13 01 PM


There must be some point in evolve where kstar_1 is reset from 4 to 1... 

I also noticed that we had to rename mass_1 to mass1_binary. This could be confusing since in bpp is is mass1_binary, but in bcm it is mass_1. Is there a way to rename the columns or is it a huge pain @scottcoughlin2014?

I'll keep playing around with stuff (including the example you have above).

michaelzevin:
@katiebreivik I think the mass1_binary is a relic of old BSE. I vote for changing it as well if we can...

katiebreivik:
It's from the initial sample, since we have some single stars and some binary stars. I'm not sure if it's a pain though :-/

As for the error I'm running into, I think it's because I tried to redefine a new single binary instead of building it out of the initC file. I'm not sure that I understand why there is a difference, but we definitely will need to add some docs here to make sure people don't make the same mistake I did.

astroJeff:
I pulled your branch, @scottcoughlin2014 and tried to test it. But for some reason I'm getting a segfault when I call Evolve.evolve(). Maybe we can work through this tomorrow? I actually have a need for this feature of cosmic.

scottcoughlin2014:
> I pulled your branch, @scottcoughlin2014 and tried to test it. But for some reason I'm getting a segfault when I call Evolve.evolve(). Maybe we can work through this tomorrow? I actually have a need for this feature of cosmic.

@astroJeff I can meet to discuss this. In fact, maybe we have a quick one off meeting to discuss this PR (as well as the naming of the `bpp` columns and the `initial conditons` columns just so we get this right. Obviously this is a big new feature and I want to make sure it is solid. @katiebreivik @michaelzevin whatcha think?

michaelzevin:
@scottcoughlin2014 fine by me, Monday or Tuesday of next week work well.

katiebreivik:
For our discussion later: I think we need to firm up the logic on how we compute the vsys values since it can get confusing if you start after one of the binary components has exploded. 
Screen Shot 2020-01-21 at 9 56 37 AM

If you restart the evolution after the first SN, you get very different values for vsys_final, which will probably cause problems when accounting for systemic velocities in populations. I know this will be important for Camille's project, but should matter for pretty much any compact object population

PR #372, opened 2020-04-16
Sana initial conditions

katiebreivik:
This PR incorporates new initial condition models for the independent sampler by request of @michaelzevin (but really everyone).

Some notes: 
- we adopt the methods of Renzo+2019 which assumes flat in log for m1 < 15 Msun and the power law for m1 >= 15 Msun. This is because the sana+2012 paper focused on *massive* stars
- the Sana+2012 observations for eccentricities are also for massive stars, which might circularize quite a bit from pre-ZAMS interactions since they are close.... thus the eccentricity is pretty severely peaked toward 0.0; for this reason, I recommend that we don't use sana12 as an ecc default for all types of stars.

tests should all be updated!

### DIFF
--- a/bin/cosmic-pop
+++ b/bin/cosmic-pop
@@ -100,6 +100,10 @@ def parse_commandline():
                         help="number of processors", type=int, default=1)
     parser.add_argument("--binary_state", nargs='+', type=int)
     parser.add_argument("--sampling_method")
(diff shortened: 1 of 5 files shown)


katiebreivik:
This won't work with cosmic-pop yet! I need to fix that up before tuesday!

Issue #357, opened 2020-04-07
Using the correct neutrino mass loss in BH formation


michaelzevin:
We need to figure out from Chris Fryer what the correct amount of neutrino mass loss to use.

michaelzevin:
After talking with Chris Fryer and Vicky, we decided it would be best to make this a flag (@katiebreivik @scottcoughlin2014 let me know if you disagree). 

BHmassloss = -1: you lose 10% of CO core mass if CO core mass is less than 5 Msun, else lose 0.1*5 = 0.5 Msun
BHmassloss \exists [0,1]: you lose CO core mass * BHmassloss

What do you think? We can also change this variable name to whatever.

michaelzevin:
This is addressed in PR #374. Will close this issue ones that PR is reviewed and merged.

michaelzevin:
Fixed in #374

PR #374, opened 2020-04-18
Remnant mass loss flag

michaelzevin:
This is addressing issue #357. 

I've changed the procedure for converting from baryonic to gravitational mass, according to recent discussions with Chris Fryer. Another important change with this PR is that mxns sets the boundary between the _final_ remnant mass rather than the baryonic remnant mass, i.e. checks whether it is a  BH or NS after the conversion from baryonic to gravitational mass.

 The new flag (`rembar_massloss`) determines the procedure for how to do this: 
- `-1 <= rembar_massloss < 0`: The prescription used in Fryer 2012 (and in Startrack), with a flexible mass-loss parameter for BHs. NSs follow Lattimer & Yahil, and BHs follow  `M_rem = M_bar,rem + rembar_massloss * M_bar,rem` (Note that since rembar_massloss is negative for this prescription, this lowers the M_rem mass. 
- `rembar_massloss >=0`: This applies the new prescription, which goes as follows: 
1. Calculate Mrem using Lattimer & Yahil
2. If Mbar,rem-Mrem > rembar_massloss, set Mrem = Mbar,rem-rembar_massloss
3. If Mrem <= mxns, kstar=13
4. If Mrem > mxns, kstar=14. Reset Mrem, do all the PPI stuff, and recalculate Mrem = Mbar,rem-rembar_massloss

I've checked this using my compact binaries mass spectrum code, and it looks to be working well. There is no more artificial mass gap between NSs and BHs!


cosmic/src/hrdiag.f
```diff
@@ -659,18 +660,26 @@ SUBROUTINE hrdiag(mass,aj,mt,tm,tn,tscls,lums,GB,zpars,
...
+     &                               mrem = mt-rembar_massloss
+                     endif
+                  endif
+
+* Determine whether a zero-age NS or BH is formed
+                  if(mrem.le.mxns)then
+                     mt = mrem
                      kw = 13
-* Convert baryonic mass to gravitational mass (Lattimer & Yahil 1989)
-                     if(nsflag.ge.2) mt = 2.108167d0*SQRT(10.d0+3.d0*mt)
```
katiebreivik, line 668: The deleted code for mt doesn't seem to match mrem on line 670. Specifically, the SQRT(10d.0 + 3.d0*mt) in 668 and the  SQRT(1.d0 + 0.3*mt) 

I'm probably missing something but this is probably worth a quick check.

cosmic/src/hrdiag.f
```diff
@@ -1013,18 +1034,26 @@ SUBROUTINE hrdiag(mass,aj,mt,tm,tn,tscls,lums,GB,zpars,
...
-* Zero-age Neutron star
-*
+* Specify the baryonic to gravitational remnant mass prescription
+* MJZ 04/2020
+
+* Determine gravitational mass using Lattimer & Yahil 1989 for nsflag>1
+                  if(nsflag.le.1)then
+                     mrem = mt
+                  else
+                     mrem = 6.6666667d0*(SQRT(1.d0+0.3d0*mt)-1.d0)
```
katiebreivik, line 1044: same check here as above at line 670

cosmic/utils.py
```diff
@@ -848,6 +848,10 @@ def error_check(BSEDict, filters=None, convergence=None, sampling=None):
     if flag in BSEDict.keys():
         if BSEDict[flag] <= 0:
             raise ValueError("'{0:s}' needs to be greater than 0 (you set it to '{1:0.2f}')".format(flag, BSEDict[flag]))
+    flag='rembar_massloss'
```
katiebreivik, line 851: I think this is currently breaking the tests

cosmic/src/hrdiag.f
michaelzevin: The full expression in the old version is:  
668                      if(nsflag.ge.2) mt = 2.108167d0*SQRT(10.d0+3.d0*mt)
669      &                                    - 6.6666667d0

which is equivalent to (but more complicated then): 
6.6666667d0*(SQRT(1.d0+0.3d0*mt)-1.d0)

cosmic/src/hrdiag.f
michaelzevin: same as comment above

cosmic/utils.py
michaelzevin: hmmm, why would this break them? any suggestions for how to  fix?

cosmic/src/hrdiag.f
katiebreivik: oh youre totally right; I was dividing instead of multiplying. sorry!

cosmic/utils.py
katiebreivik: you can see how they broke here: https://travis-ci.com/github/COSMIC-PopSynth/COSMIC/jobs/320663446

it looks like the test_single_evolve_with_inifile and test_multi_evolve_with_inifile are having some trouble with the inifile. I think @scottcoughlin2014 wrote these tests? Maybe they just need to be updated? It looks like the indices are hard coded in the test.

PR #381, opened 2020-04-28
Init c bugs

katiebreivik:
the power law sampler had incorrect documentation, so that's now fixed. I've also updated the sana12 eccentricity sampler.


katiebreivik:
I also carefully checked the initial conditions filter for the independent sampler. The result of 

`import matplotlib.pyplot as plt`
`from cosmic.sample.sampler import independent`
`initconditions = independent.Sample()`
`mass1, total_mass1 = initconditions.sample_primary(primary_model='kroupa01', size=100000)`
`mass2 = initconditions.sample_secondary(mass1)`
`plt.hist(mass2/mass1)`
`plt.show()`

produces [mass_ratio_ind.pdf](https://github.com/COSMIC-PopSynth/COSMIC/files/4546826/mass_ratio_ind.pdf)

after filtering with:
`mass1_binaries, mass_single, binfrac_binaries = initconditions.binary_select(mass1, binfrac_model=1.0)`
`mass2_binaries = initconditions.sample_secondary(mass1_binaries)`
the plot stays the same

however, after filtering with:
`final_kstar = [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14]`
`primary_min, primary_max, secondary_min, secondary_max = mass_min_max_select(final_kstar, final_kstar)`
`ind_select_primary, = np.where((mass1_binaries > primary_min) & (mass1_binaries < primary_max))`
`ind_select_secondary, = np.where((mass2_binaries > secondary_min) & (mass2_binaries < secondary_max))`
`ind_select = list(set(ind_select_primary).intersection(ind_select_secondary))`
`mass1_binary = mass1_binaries[ind_select]`
`mass2_binary = mass2_binaries[ind_select]`
`plt.hist(mass2_binary/mass1_binary)`
`plt.show()`

the plot becomes: [mass_ratio_ind_filter.pdf](https://github.com/COSMIC-PopSynth/COSMIC/files/4546851/mass_ratio_ind_filter.pdf)

At first glance, this all seems wrong. BUT I don't think it is. In the sampler, we account for the total mass sampled at the phase of mass1_binaries and mass2_binaries. Then we filter on m1 > 0.08, m2 > 0.08 (or whatever the final kstars cause the filter to be), then pass back the filtered mass1_binary and mass2_binary. Since we are accounting for the mass before filtering I *think* we are ok. We should talk about this though.

michaelzevin:
@katiebreivik per your comment above, did we come to a consensus about whether this interpretation is correct or not? Or do we need to talk more?

katiebreivik:
I think we sorted it out on the call. We now have the interpretation:
both m1 and m2 are limited to be above 0.08, but m1 is limited to between [0.08, 150] and m2 is limited to [0.08, m1]. The effect of the m2 limit is seen here: 
[q_from_M2_lims.pdf](https://github.com/COSMIC-PopSynth/COSMIC/files/4549284/q_from_M2_lims.pdf)

cosmic/sample/sampler/independent.py
```diff
@@ -408,12 +409,12 @@ def sample_porb(self, mass1, mass2, ecc, model='sana12', size=None):
             porb = porb_yr*yr_day
         elif model == 'sana12':
             from cosmic.utils import rndm
-            porb = 10**rndm(a=0.15, b=5.5, g=0.55, size=size)
+            porb = 10**rndm(a=0.15, b=5.5, g=0.45, size=size)
```
michaelzevin, line 412: Are you sure about this? From the supplemental materials of [Sana+2012](https://science.sciencemag.org/content/sci/suppl/2012/07/25/337.6093.444.DC1/1223344.Sana.SM.pdf), it indeed looks like the exponent for the orbital period power law (\pi) is -0.55

cosmic/sample/sampler/independent.py
```diff
@@ -408,12 +409,12 @@ def sample_porb(self, mass1, mass2, ecc, model='sana12', size=None):
...
         elif model == 'renzo19':
             from cosmic.utils import rndm
             porb = 10**(np.random.uniform(0.15, 5.5, size))
             ind_massive, = np.where(mass1 > 15)
-            porb[ind_massive] = 10**rndm(a=0.15, b=5.5, g=0.55, size=len(ind_massive))
+            porb[ind_massive] = 10**rndm(a=0.15, b=5.5, g=0.45, size=len(ind_massive))
```
michaelzevin, line 417: Same as comment above. Maybe you're mixing this up with the eccentricity power law (\eta =  0.45)?

cosmic/sample/sampler/independent.py
michaelzevin: Oh I see...If `g` is indeed 1+x where x is the powerlaw exponent, then we're gucci. 

But to be honest, for the sake of clarity, why don't we just feed in the powerlaw exponent directly into rndm and add 1 inside of that function?

cosmic/tests/test_sample.py
```diff
@@ -9,33 +9,73 @@
...
 PORB = 670.88711347
 PORB_RENZO19 = 303.789477175485
-PORB_SANA12 = 43.52635793101753
-THERMAL_ECC_SUM = 5.7488819291685695
-UNIFORM_ECC_SUM = 3.58801017672414
-SANA12_ECC_SUM = 0.03346470981696298
+SANA12_PORB_POWER_LAW = -0.55
+FLAT_SLOPE = 0.0
+THERMAL_SLOPE = 2.0
+SANA12_ECC_POWER_LAW = -0.42
```
michaelzevin, line 34: We should either stick the best-fit parent distribution numbers (porb:-0.55, ecc:-0.45) or the mean of the retrieved  parameters (porb:-0.38, ecc=-0.42). But not mix and match. 

I'd recommend the first set.

cosmic/utils.py
```diff
@@ -514,7 +514,7 @@ def newton_cotes(x, f) :
 
 
 def rndm(a, b, g, size):
-    """Power-law generator for pdf(x)\propto x^{g} for a<=x<=b
+    """Power-law generator for pdf(x)\propto x^{g-1} for a<=x<=b
```
michaelzevin, line 517: again, I think things would be much clear throughout the code if  rndm takes in the powerlaw exponent directly

cosmic/tests/test_sample.py
```diff
@@ -88,34 +159,51 @@ def test_binary_fraction(self):
...
         np.random.seed(2)
-        porb = SAMPLECLASS.sample_porb(1.0, 1.0, 0.5, 'renzo19', size=1)
-        self.assertAlmostEqual(porb[0], PORB_RENZO19)
+        # Check that the sample_porb function samples porb properly
+        porb = SAMPLECLASS.sample_porb(mass1, mass2, ecc, 'log_uniform', size=len(mass1))
+        # filter out the weird cuts for RLO and convert to sep
+        sep = a_from_p(porb, mass1, mass2)
+        sep = sep[sep > 10]
+        uniform_slope = linear_fit(np.log10(sep))
+        self.assertEqual(np.round(uniform_slope, 1), FLAT_SLOPE)
```
michaelzevin, line 206: Haven't gone through all these tests in much detail, but I trust that they're giving the desired results

cosmic/sample/sampler/independent.py
```diff
@@ -208,8 +208,8 @@ def sample_primary(self, primary_model='kroupa01', size=None):
             multiplier = 1
             a_0 = np.random.uniform(0.0, 1, size)
 
-            low_cutoff = 0.740074
-            high_cutoff = 0.908422
+            low_cutoff = 0.771
+            high_cutoff = 0.919
```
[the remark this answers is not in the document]
katiebreivik: If you have a broken power law, you need to pick the relative contribution of each of them. The cutoffs set the normalization of what fraction of stars are distributed according to each power law.  The numbers come from Hurley+2002 Eqs 95-96. You get a similar cutoff for kroupa01. 

You can also verify that this produces the right behavior because the histograms don't have big jumps at the masses where the power law switches:

primary_mass_kroupa93
primary_mass_kroupa_01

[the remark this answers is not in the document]
katiebreivik: the top one is kroupa93 and the bottom one is kroupa01. The differences are small, but verifiable by the tests.

cosmic/sample/sampler/independent.py
```diff
@@ -224,20 +224,20 @@ def sample_primary(self, primary_model='kroupa01', size=None):
...
+            # Since COSMIC/BSE can't handle < 0.08Msun, we will truncate
+            # at 0.08 Msun instead of 0.01
+
             total_sampled_mass = 0
             multiplier = 1
             a_0 = np.random.uniform(0.0, 1, size)
 
-            low_cutoff = 0.37148816884988606
-            high_cutoff = 0.8496015751162523
+            cutoff = 0.748
```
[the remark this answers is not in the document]
katiebreivik: see the above set of comments.

cosmic/sample/sampler/independent.py
katiebreivik: yeah. I'll fix the rndm method in utils. I agree that this makes the most sense since future users won't have to make the same 'omg what is a power law' journey I made yesterday.

katiebreivik:
Lastly - @scottcoughlin2014 Here is a jupyter notebook that shows how these things work and what their histograms are: 
[independent_initital_conditions.ipynb.zip](https://github.com/COSMIC-PopSynth/COSMIC/files/4551714/independent_initital_conditions.ipynb.zip)

cosmic/sample/sampler/independent.py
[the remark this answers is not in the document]
michaelzevin: ok...why don't the  cutoffs normalize to unity then? Sorry for not being used to how broken power laws are generated :)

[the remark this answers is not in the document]
michaelzevin: nevermind, I  see that the distribution integrates to unity

cosmic/sample/sampler/independent.py
```diff
@@ -455,7 +456,7 @@ def sample_ecc(self, ecc_model='thermal', size=None):
             ecc = rndm(a=0.001, b=0.9, g=-0.42, size=size)
```
michaelzevin, line 456: Still wondering about the choice of eccentricity here. Why are we using -0.42  as  default rather  than -0.45?

katiebreivik: I'm following eq 3 of De Mink & Belczynski 2015: https://iopscience.iop.org/article/10.1088/0004-637X/814/1/58/pdf

michaelzevin: If we want to adopt their values, then I think we should do the same for the orbital period (Sana's  best fit is 0.55, they use 0.5 for their standard assumption). I think it's clear to just use the best-fit values from Sana, even though there is a spread  for what could  be considered viable. What do you think?

katiebreivik: I think the best fit for the eccentricity is -0.42. I'm looking at page 8 of the supplement in the second paragraph. eta is the power law that they fit for eccentricity and pi is the power law for orbital period.

michaelzevin: From the Sana paper,  pg 8: 

"We generate 40 synthetic populations that share the properties of our sample
and are randomly drawn from the best-fit parent distributions, i.e. fbin = 0.69, π = -0.55, κ
= -0.10, and η = -0.45. Each synthetic population is input to our code and we re-derive
the best-fit intrinsic binary fraction and parent distributions. The mean and 1σ dispersion
of the retrieved parameters serve as a check for the presence of systematic biases and as a
probe for the accuracy of the procedure. We obtain <fbin> = 0.74 ± 0.09, <π> = -0.38 ±
0.22, <κ> = -0.21 ± 0.58 and <η> = -0.42 ± 0.17. "

The best-fit parameters are the first numbers,  the recovered parameters from their 40 realizations are the second set of numbers. They're trying (poorly) to prove that there aren't systematic  biases. But the best-fit parameters  are still the top set of numbers.

katiebreivik: ah! So I was misunderstanding. I think the De Mink & Belczynski paper has this wrong them as well. I really should have taken a stats class 🙃

I'll fix this right away!

katiebreivik: Thanks for sticking this out @michaelzevin -

katiebreivik:
@scottcoughlin2014 this is ready for you to check over

Issue #368, opened 2020-04-14
Update ini file with new values


michaelzevin:
such as gamma=-2, etc.

michaelzevin:
Addressed in #385, though we will need to double check these as we go  through the new docs

PR #385, opened 2020-05-08
updated the new default values for the Params.ini file. In the proces…

michaelzevin:
…s, we also added more info about the magnetic field evolution parameters, and  set their default  values to positive rather than negative. This is  related to PR #368


michaelzevin:
@scottcoughlin2014 it is ok if I merge this?

katiebreivik:
This looks good to go

Issue #382, opened 2020-04-28
Change WR winds to actually match Hamann paper and add correct citations


michaelzevin:
http://articles.adsabs.harvard.edu/pdf/1995A&A...299..151H
https://arxiv.org/pdf/1102.0530.pdf
https://www.aanda.org/articles/aa/pdf/2006/46/aa5912-06.pdf
https://www.aanda.org/articles/aa/pdf/2005/44/aa4030-05.pdf
https://www.aanda.org/articles/aa/pdf/2005/41/aa2862-05.pdf

michaelzevin:
Taken care of in #386

PR #386, opened 2020-05-08
did  some checks and updated documentation in the mlwinnd  function. …

michaelzevin:
…Removed the ability to use eddlimflag for prescriptions not used in Giacobbo et al. 2018


michaelzevin:
Addresses #382

Issue #388, opened 2020-05-09
need to update docs for running cosmic-pop


katiebreivik:
need to include kick info in description of DataFrames saved in the dat files

Issue #375, opened 2020-04-20
bpp doesn't include effective temperature which


katiebreivik:
This would be a really helpful addition to the bpp so that we can do HR-diagram type filtering now that we are nearing Scotty's PR #367 being finished and able to willy nilly control our timesteps.

Issue #356, opened 2020-04-07
B0_1 does not change in bpp


michaelzevin:
The B0_1 and B0_2 parameters in the bpp array do not change, though they do in  the bcm array. We should have the *current*  values of the B-field in the bpp array.

scottcoughlin2014:
related to #354

michaelzevin:
Also, since in both the bcm and bpp B0 will be the _current_ B-field, we should change these column names to "B_1" and "B_2".

Issue #354, opened 2020-04-02
B_0_1 documentation incorrect


michaelzevin:
Small thing, but just wanted to make sure this gets fixed before the next release. The parameter B_0_1 is described wrong on the docs page (https://cosmic-popsynth.github.io/docs/stable/output_info/index.html). It says that it is the "Initial neutron star magnetic field [{\rm{G}}]" but Claire said that in reality this parameter is actually the "current magnetic field".

katiebreivik:
Good catch!

michaelzevin:
Also, this should be changed from B_0_1 > B0_1 in the documentation

PR #389, opened 2020-05-11
Bpp array fixes

katiebreivik:
fixes for #388, #375, #356, #354


scottcoughlin2014:
Looks good as soon as we get that initial conditions for testing fixed

katiebreivik:
I can't for the life of me figure out what is happening with this file..... I know that the test works locally, but can't figure out why we would have a merge conflict?

PR #395, opened 2020-05-14
Periastron check

astroJeff:
Addresses bug related to mass transfer in eccentric binaries, when ST_tide=0. Previously the Roche lobe was calculated assuming circular binaries only, but of course in eccentric binaries, RLO can happen at periastron. Furthermore, there was no setting to cause instant circularization upon RLO. 

Now, when ST_tide=0 (and 1), the binary starts mass transfer and circularized the orbit as soon as it overfills its Roche lobe at periastron.


astroJeff:
This PR also accidentally includes a hot fix for COSMIC's plotting script.

Issue #396, opened 2020-05-14
(Thermal timescale?) mass transfer issue


astroJeff:
When a massive binary enters RLO, if the accretor is a main sequence star, it seems to be able to accept all of the donated mass. In one particular case, I'm finding 30 Msun is (nearly completely) transferred onto a 35 Msun star over 1500 years. This corresponds to a mass transfer rate of 0.02 Msun/yr, and is almost certainly unphysical. We need to look deeper into how COSMIC deals with thermal timescale mass transfer.

astroJeff:
I looked into this a bit more. From the BSE paper, they calculate the Kelvin-Helmholtz timescale (without reference) as 10^7 M/RL, as opposed to GM^2/RL. In the particular case I was looking, the BSE version is ~2 orders of magnitude shorter than what I'm familiar with. The result is that the star can accept much more mass than I think it ought to be able to.

katiebreivik:
Thanks for checking the paper - I was just about to head to that. It's definitely bad that they do this scaling (and it looks like it fits *ok* to within a factor of 3 for solar mass stars). I can cook up a branch that checks how much this changes. I'll also check the other timescales to be sure we don't have other scalings.

katiebreivik:
oh wait! the Timescale is not quite 10^7 M/RL in BSE it's a bit more complicated:
![Screen Shot 2020-05-14 at 10 45 36 AM](https://user-images.githubusercontent.com/22921809/81955751-1a961100-95d0-11ea-84fb-7df6c0add89c.png)

Comparing to a 1 Msun MS star:
t_KH_BSE = 1e7 * (1) * (1) / (1 * 1) = 1e7
t_KH_true = G * M^2/R*L = 31.3e6
so the difference is ~3

for a 30 Msun star:
t_KH_BSE = 1e7 * (30) * (30) / (9.71 * 2e5) ~ 4634.4 yr
t_KH_true = G * M^2/R*L ~14505.7 yr
difference ~ 3

astroJeff:
It looks like COSMIC does in fact multiply by the extra mass term (or m-m_env). My mistake! As @katiebreivik has said, it's only a factor of ~3 off from the term I'm familiar with.

So maybe the issue is a logical one. I still don't understand how/believe a main sequence star can accept mass at such a high rate. Certainly this is not what StarTrack does. I will keep looking into this.

katiebreivik:
[...]

PR #397, opened 2020-05-14
Thermal mt fix

astroJeff:
Lowers mass transfer rate allowed to be accepted by the accretor for thermal timescale mass transfer down to 1x the thermal rate from 10x the thermal rate.

Resolution for Issue #396


michaelzevin:
Hey everyone, after talking with Tassos I think we need to be more careful before changing this. I'm doing some literature diving, but I'll post more info in Slack.

Issue #175, opened 2019-04-23
Possible updates to COSMIC's binary evolution


michaelzevin:
- [x] Bray & Eldridge kick prescriptions

- [ ] Nanging lambda

- [ ] Clausen supernova prescriptions

- [x] Ultra-stripped supernovae, bimodal kick distribution or reduced natal kicks?

- [ ] Dynamical stability of mass transfer? See Vigna-Gomez et al. 2018, section 2.2.4

- [x] Mass transfer stability criteria (always allowing for table Case BB mass transfer, might replace the Tauris fitting formulae)

- [ ] ECSN remnant mass?

PR #399, opened 2020-05-14
Added new flag called kickflag, that allows users to drawn natal kick…

michaelzevin:
…s from prescriptions in Bray & Eldridge 2016 and Giacobbo & Mapelli 2020. This tackles one of the points  in #175


michaelzevin:
@scottcoughlin2014 not sure why the latest commit failed, as I just changed things back in the example Params.ini file

cosmic/src/kick.f
```diff
@@ -87,6 +88,13 @@ SUBROUTINE kick(kw,m1,m1n,m2,ecc,sep,jorb,vk,snstar,
 * radius in solar units.
       gmrkm = 1.906125d+05
 
+* Set values for mean NS mass and mean ejecta as in Giacobbo & Mapelli 2020
```
katiebreivik, line 91: Is there any reason we might want to have these as variables that the user puts in? I'm most asking because I wonder if the values will change once we have more NS binary observations that change the merger rates?

cosmic/src/kick.f
```diff
@@ -191,18 +201,30 @@ SUBROUTINE kick(kw,m1,m1n,m2,ecc,sep,jorb,vk,snstar,
...
+                vk = 0.d0
+             elseif(kw.eq.14.and.bhflag.eq.1)then
+                 fallback = MIN(fallback,1.d0)
+                 vk = MAX((1.d0-fallback)*vk,0.d0)
+                 vk2 = vk*vk
+             elseif(kw.eq.14.and.bhflag.eq.2)then
+                vk = vk * mxns / m1n
+                vk2 = vk*vk
+             endif
+          elseif(kickflag.eq.-1)then
```
katiebreivik, line 217: Does kickflag = [-1, -2] rely on remnantflag to be >= 3  since the remnant masses derived for mean_mns depend on the Fryer 2012 models? (That's my takeaway from the Giacobbo and Mapelli paper). If this is the case, we should throw an error if remnant flag < 3.

michaelzevin: @katiebreivik that's not a bad idea. One we should also consider is increasing the proto-core mass to 1.2. This isn't done in Fryer 2012, but Mario said that Chris recommended doing this to match NS masses more. Thoughts?

cosmic/src/kick.f
michaelzevin: I thought about this, but I worry that this will introduce way too many flags. They're easy enough to change in the code directly, since they're set right at the beginning of kick.f. Thoughts?

cosmic/src/kick.f
michaelzevin: On second thought, I think it would be best not to raise an error. This won't really affect things, and remnant flags < 3 probably won't be used anyways. The assumption of the mean mns can be independent from the choice of kicks. 

However, I'm asking Mario about changing mean_mns from 1.0 to 1.1. See left column of https://arxiv.org/pdf/1909.06385.pdf page 3, last paragraph.

PR #402, opened 2020-05-15
Changed natal kick array to be set by primary/secondary

### LINKED ISSUE #398: Controlling the kick of a specific star
Currently, the natal_kick_array only allows us to control the kick of the first/second SN and not the kick associated with the stars. Has the option to directly control the kicks of the stars been considered?

michaelzevin:
Addressing issue #398

changed natal kick array to have the rows indicate the primary and secondary rather than the first and second star to go SN. Also fixed documentation for this input.


scottcoughlin2014 (approved):
looks good now

PR #404, opened 2020-05-15
Compas qcrits

michaelzevin:
added 5th qcflag to match COMPAS, so now we have BSE, StarTrack, binary_c, and COMPAS. 

These are directly out of Neijssel+2020 (https://ui.adsabs.harvard.edu/abs/2019MNRAS.490.3740N/abstract)

The MS and HG qcrits are calculated by converting from the fixed zetas to qcrits, which is what Simone did. These are fairly straightforward to calculate, especially with Mathematica. Just set the below values to dlog(R_RLO)/dlog(m), where R_RLO is the appoximation from Eggleton 1983
MS:  dlog(R∗)/dlog(m) = 2.0
HG:  dlog(R∗)/dlog(m) = 6.5

Here's also a table we can use for docs or talking with vicky or whatever

[qcrit_table.pdf](https://github.com/COSMIC-PopSynth/COSMIC/files/4637315/qcrit_table.pdf)

### DIFF
--- a/cosmic/src/evolv2.f
+++ b/cosmic/src/evolv2.f
@@ -2046,7 +2046,7 @@ component.
       elseif(qcflag.eq.3)then
 *
 * Use the binary_c prescriptions taken from Claeys+2014 Table 2
-* but w/ Hjellming & Webbing for GB/AGB
+* but w/ Hjellming & Webbink for GB/AGB
 * If q1 = m_donor/m_acc > qc then common envelope
 *
          if(kstar(j2).lt.10)then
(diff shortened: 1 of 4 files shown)


michaelzevin:
FYI, if you want to check the zeta to qcrit conversion, you  can use this snippet of  code: 

```from scipy.optimize import fsolve
def f(q,dlnRdlnm):
    return 4*q - 2./3*(1+q) * ( 2 + (0.6*q**(2./3) + 0.5*q**(1./3)*(1+q**(1./3))**(-1.))/(0.6*q**(2./3)+np.log(1+q**(1./3)))) - dlnRdlnm
print('qcrit for MS stars (dlnRdlnm = 2.0) : ', round(fsolve(f, 3., args=2.0)[0],3))
print('qcrit for HG stars (dlnRdlnm = 6.5) : ', round(fsolve(f, 3., args=6.5)[0],3))
#qcrit for MS stars (dlnRdlnm = 2.0) :  1.717
#qcrit for HG stars (dlnRdlnm = 6.5) :  3.825```

File: cosmic/src/evolv2.f [hunk 1]
```diff
@@ -2046,7 +2046,7 @@ SUBROUTINE evolv2(kstar,mass,tb,ecc,z,tphysf,
       elseif(qcflag.eq.3)then
 *
 * Use the binary_c prescriptions taken from Claeys+2014 Table 2
-* but w/ Hjellming & Webbing for GB/AGB
```
katiebreivik, line 2049: nice catch haha

katiebreivik (commented):
this looks good - I think we should wait for the more complicated merges though just to keep things simple.

katiebreivik (approved):

PR #407, opened 2020-05-18
MT updates

katiebreivik:
This PR adds three flags:
f_a : to be added next
don_lim : limits the amount of mass the donor can lose
acc_lim : limits the amount of mass the accretor can handle from the donor


katiebreivik:
f_acc is applied after the don_lim and acc_lim models are applied - I *think* this is the right way to do this but we should definitely talk about this tomorrow to set it in stone

cosmic/src/evolv2.f
```diff
@@ -2497,27 +2497,12 @@ SUBROUTINE evolv2(kstar,mass,tb,ecc,z,tphysf,
...
-*           if(kstar(j1).eq.2)then
-*              mew = rad(j1)/rol(j1) - 1.d0
-*              mew = 2.d0*mew
-*              dm1 = dm1*10.d0**mew
-*           endif
-***
-            dm1 = MIN(dm1,mass(j1)*tb/tkh(j1))
-         elseif(rad(j1).gt.10.d0*rol(j1).or.(kstar(j1).le.1.and.
+         if(rad(j1).gt.10.d0*rol(j1).or.(kstar(j1).le.1.and.
      &          kstar(j2).le.1.and.q(j1).gt.qc))then
```
katiebreivik, line 2506: check that qc stuff is already taken care of by qcrits above

cosmic/src/evolv2.f
```diff
@@ -2541,9 +2526,27 @@ SUBROUTINE evolv2(kstar,mass,tb,ecc,z,tphysf,
...
+* Now treat stable mass transfer through RLO
+*
+
+*
+* KB: choose the timescale to limit donor mass loss rate
+*
+
+         if(don_lim.eq.0)then
+            if(kstar(j1).ge.2.and.kstar(j1).le.9.and.kstar(j1).ne.7)then
+               t_don_lim = tkh(j1)
```
katiebreivik, line 2541: we are unclear on why MS has tdyn while HG, GS have tkh

cosmic/src/evolv2.f
```diff
@@ -2541,9 +2526,27 @@ SUBROUTINE evolv2(kstar,mass,tb,ecc,z,tphysf,
...
+            else
+               t_don_lim = tdyn
+            endif
          else
```
katiebreivik, line 2545: make elseif(don_lim.eq.-1)then

cosmic/src/evolv2.f
```diff
@@ -2696,8 +2723,14 @@ SUBROUTINE evolv2(kstar,mass,tb,ecc,z,tphysf,
 *
 * We have a giant whose envelope can absorb any transferred material.
 *
-            dm2 = dm1
+            dm2 = MIN(1.d0,t_acc_lim)*dm1
```
katiebreivik, line 2726: check if ST allows full acc onto giants

katiebreivik: just confirming here - according to both Belczysnki+2008 and Dominik+2013, ST does not limit accretion onto the donor based on any thermal responses -- the accretion fraction is just applied as f_acc = 0.5. This is justified by a citation to Dewi & Tauris 2003 that says that massive stars accrete between 40%-70% of the donor's mass loss.

cosmic/src/evolv2.f
katiebreivik: confirming here that we use this to check to see if main sequence donor will merge with accretor.

cosmic/src/evolv2.f
katiebreivik: done!

scottcoughlin2014:
Whenever we fix this it will take on a different look than this PR

PR #410, opened 2020-05-19
Fix to ECSN kicks with newly-added kick prescriptons

michaelzevin:
Fixed small bug with sigmadiv so that ECSN can be tracked even when sigmadiv is negative (i.e., the user gives a set value for the ECSN sigma). We were doing this incorrectly before. 

Also now pass sigmahold to comenv and kick, so that we know what the original input for the sigma parameter is and we can revert to it when using certain kick distributions.

### DIFF
--- a/cosmic/src/kick.f
+++ b/cosmic/src/kick.f
@@ -1,6 +1,6 @@
 ***
       SUBROUTINE kick(kw,m1,m1n,m2,ecc,sep,jorb,vk,snstar,
-     &                r2,fallback,kick_info,disrupt)
+     &                r2,fallback,sigmahold,kick_info,disrupt)
       IMPLICIT NONE
       INCLUDE 'const_bse.h'
 *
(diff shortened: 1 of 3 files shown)


michaelzevin:
Looks like it passed other than the coveralls

katiebreivik (approved):
looks good - thanks @michaelzevin

cosmic/src/kick.f
```diff
@@ -230,8 +230,8 @@ SUBROUTINE kick(kw,m1,m1n,m2,ecc,sep,jorb,vk,snstar,
 * Use kick scaling from Giacobbo & Mapelli 2020, Eq. 2
              vk = vk * ((m1-m1n)/mean_mej)
           elseif(kickflag.eq.-3)then
-* Use kick scaling from Eldridge & Stanway 2016, Eq. 1
-             vk = alphakick * (m1/m1n) + betakick
+* Use kick scaling from Bray & Eldridge 2016, Eq. 1
+             vk = alphakick * ((m1-m1n)/m1n) + betakick
```
katiebreivik, line 234: good catch here!

katiebreivik (approved):

PR #416, opened 2020-06-25
Timestep conditions hotfix

katiebreivik:
This is a tiny PR that updates the list of timestep conditions that are allowed because we forgot about changing rrlo in the list. This should probably be merged into 3.3 quickly.


katiebreivik:
I don't understand why my commits are so messed up here .... sorry!

PR #422, opened 2020-08-11
Adding wind prescriptions from Beasor 2020

eeshadg:
I added windflags 4 and 5 based on Beasor et al 2020. These flags explicitly decrease wind mass loss for red supergiants. Windflag 4 applies it for all masses and windflag 5 only applies it for initial masses 8-25 Msun (specified in Beasor et al 2020)


katiebreivik:
Hey @eeshadg - I was looking over this and found that we forgot to add in the update to cosmic/utils.py that tells the error checker to accept windflag = 4,5. Can you update that and push the changed utils.py file as well?

katiebreivik:
I'm closing this for now and we will re-open it once we have the branch all together.

PR #423, opened 2020-08-12
Grflag addition

astroJeff:
Added a flag to turn on/off orbital decay due to gravitational wave radiation


cplb:
Does this work outside of COSMIC? It would make my life so much easier

astroJeff:
What do you mean outside of COSMIC?

cplb:
The Universe in general

scottcoughlin2014:
@astroJeff how does it feel to have become a COSMIC BSE flag adding pro? it is ok if the answer is not great haha, I will review this on Tuesday!

astroJeff:
Yeah... I don't think I can call myself a pro yet. You'll want to check the code to make sure I didn't screw anything up.

Issue #425, opened 2020-08-24
Initial mass cutoffs are bad if you want to look at bright stars and not compact objects


katiebreivik:
If you, e.g., want to get BH + Giant star companions, you would specify an initial population as:

`final_kstar1 = [14]
final_kstar2 = [2,3,4,5,6,]
initBin_n, m_singles_n, m_binaries_n, n_singles_n, n_binaries_n = InitialBinaryTable.sampler('independent', final_kstar1, final_kstar2, binfrac_model=0.5, primary_model='kroupa01', ecc_model='sana12', porb_model='sana12', SF_start=13700.0, SF_duration=0.0, met=0.02, size=10000)`

this produces:
![BH_GS_cut](https://user-images.githubusercontent.com/22921809/91053427-c9df7c80-e5f0-11ea-8bab-7a6b48e857fb.png)

However, what you really want must be specified different to get the full range of possible bright stars, especially if you are using a non-delta burst star formation history:
`final_kstar1 = [14]
final_kstar2 = [2,3,4,5,6,7,8,9,10,11,12,13,14]
initBin_n, m_singles_n, m_binaries_n, n_singles_n, n_binaries_n = InitialBinaryTable.sampler('independent', final_kstar1, final_kstar2, binfrac_model=0.5, primary_model='kroupa01', ecc_model='sana12', porb_model='sana12', SF_start=13700.0, SF_duration=0.0, met=0.02, size=10000)`

this produces:
![BH_GS](https://user-images.githubusercontent.com/22921809/91053555-fa271b00-e5f0-11ea-8429-5f8c9692508a.png)

I think the solution is to only place mass cuts for kstars in [10,11,12,13,14] and nothing else.

katiebreivik:
solved!

PR #426, opened 2020-08-24
Mass filtering

katiebreivik:
Fixes the upper mass filtering problem for kstar <= 10 referenced in Issue #425


katiebreivik:
OOHHH NOOOO - This somehow pulled in all the FIRE stuff that we don't want in the general repository. @scottcoughlin2014 -- what do I do to pull out everything pre git hash:  0281e26? It looks like I merged high_res_MW into develop a while ago...

PR #428, opened 2020-08-25
Kick f fix

astroJeff:
Kick.f changes: eccentric anomaly to mean anomaly, degrees-to-radian bugs, documentation update


scottcoughlin2014:
@astroJeff  I think the change to the kick_info will make the unit tests as constructed fail (not because it is wrong but because the unit tests were based on that bug. I will make a PR to your branch to update the initC for the tests

PR #453, opened 2020-10-27
Generic powerlaw try 2

scottcoughlin2014:


katiebreivik:
@scottcoughlin2014 -- I have verified that this works for everything except if the alpha is set to 1.0. This is the same problem I think we ran into before where a divide by zero happens since the dividend from the last line of utils.rndm is: 1 + g = 0.0  where from the updated code g = -1 * alpha = -1.

Maybe the best option here is to just catch the special case for alpha=1.0 and not use rndm. Happy to hear other suggestions though!

PR #455, opened 2020-10-27
Evolve singles

scottcoughlin2014:


katiebreivik (approved):
This looks great and works well! Thanks @scottcoughlin2014

katiebreivik:
i just realized -- we should include this as a flag in the cosmic-pop runs and in Params.ini.

scottcoughlin2014:
@katiebreivik ya good point I will do this

PR #468, opened 2020-12-01
Making thermal timescale limits apply to main sequence stars as well as giants

scottcoughlin2014:


katiebreivik (approved):
This was confirmed to do exactly what we want it to do and so should be merged!

PR #477, opened 2021-01-30
Fixed Samplers

carlrodriguez:
Ok both King and Elson profiles now work and agree with CMC.  King generator is actually *faster* than C code for large clusters.  Also fixed a few typos, created a new sampler (cmc_point_mass) which only does N-body stuff with no stars, and added r_max sampling for binaries (i.e. independent sampler will only sample binaries out to the hard/soft boundary of the cluster).

### DIFF
--- a/cosmic/sample/sampler/cmc.py
+++ b/cosmic/sample/sampler/cmc.py
@@ -114,29 +114,6 @@ def get_cmc_sampler(
     mass_binaries += np.sum(mass1_binaries)
     mass_binaries += np.sum(mass2_binaries)
 
-    # track the total number sampled
-    n_singles += len(mass_single)
-    n_binaries += len(mass1_binaries)
-
-    # select out the primaries and secondaries that will produce the final kstars
-    mass1_binary = np.array(mass1_binaries)
-    mass2_binary = np.array(mass2_binaries)
-    ecc = initconditions.sample_ecc(ecc_model, size=mass1_binary.size)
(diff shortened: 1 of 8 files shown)


cosmic/sample/sampler/cmc.py
```diff
@@ -114,57 +105,55 @@ def get_cmc_sampler(
...
-    # set singles id
-    single_ids = np.arange(mass1.size)
-    binary_secondary_object_id = np.arange(mass1.size, mass1.size + mass2_binaries.size)
-
-    # set binind and correct masses of binaries
-    binind = np.zeros(mass1.size)
-    binind[binary_index] = np.arange(len(binary_index)) + 1
-    mass1[binary_index] += mass2_binary
+    Reff2 = initconditions.set_reff(mass2_binaries, metallicity=met, **kwargs)
+    ## TODO: CARL!!! MAKE GODDAMN SURE THESE ARE ALL IN RSUN AND NOT AU!!!!!
```
katiebreivik, line 135: Has this been made sure?

scottcoughlin2014: I think that We covered this today and we await Carls final tweaks!

katiebreivik (approved):
Ok -- this all looks good and I think we are ready to merge it in. One quick note to @carlrodriguez -- the Docs string format is set by numpy doc, so when we build the docs, they won't look quite the same as the cosmic initial sampler modules. Probably worth updating eventually.

PR #478, opened 2021-02-02
sub-Eddington accretion in the docs and example Params.ini

michaelzevin:
changed docs to reflect that eddfac flag can be less than 1, which allows for sub-Eddington accretion. The flag checks in utils.py already allow for any eddfac >= 0. 

Let me know if I missed any docs!


katiebreivik (commented):
looks good! Thanks @michaelzevin

katiebreivik:
Not sure why we hadn't merged this in yet!

PR #482, opened 2021-02-15
found bug in utils for filtering bcm and bpp

### LINKED ISSUE #484: We need to allow a choose your own adventure filtering on what types of systems/evolutions you want to keep

katiebreivik:
Jared Siegel pointed out that we are missing a significant amount of LMXBs because we are filtering on the final kstars in the bcm array but an LXMB won't necessarily have an LMXB final kstar at the end of it's evolution.


astroJeff:
When Jared mentioned this issue he found, I was wondering what the problem was. Glad it's a simple fix.

astroJeff (approved):

scottcoughlin2014:
filtering needs total over haul

PR #485, opened 2021-02-24
Fix to bug with acc_lim flag not imposing Eddington limit

### LINKED ISSUE #467: Mass loss/gain rates seem much too high
Here are some initial conditions which give binaries that experience a common envelope with the primary star as the donor, then a stable mass transfer from the secondary which occurs long enough that the NS which forms from the primary accretes enough mass to collapse into a BH.
[thermal_MT_test.h5.zip](https://github.com/COSMIC-PopSynth/COSMIC/files/5616444/thermal_MT_test.h5.zip)

So -- the issue with our thermally unstable, but dynamically stable mass transfer prescription is that the current code effectively never applies a thermal timescale limit for the donor star's mass loss. We calculate the total amount of mass lost following the Hurley+2002 equation starting at this line: https://github.com/COSMIC-PopSynth/COSMIC/blob/a0ab5ad8df135fd24020fe629780f1e6aa160950/cosmic/src/evolv2.f#L2496

Screen Shot 2020-11-30 at 9 24 39 AM

The M_dot is turned into a dm as: dm = m_dot * dt where dt = porb. In a pdb, you can show that the dm is of order ~1e-8 - 1e-5 Msun, while the thermal dm: dm_therm = M_donor / thermal_timescale * porb is of order ~1e-2 - 1e-1. This means that the thermal timescale limit is not being applied.

I think a good place to start is figuring out how to set a gdb conditional statement where we can track the size of dm and/or the timesteps.

michaelzevin:
hopefully fixed the bug that doesn't account for eddington when acc_lim>0 is set. I just moved this outside the large if statement, was concerned that this wouldn't work because of the white dwarf accretor with kstar leq 6 part of the if statement, but after reading through it a few times I'm pretty sure it should be fine (since the eddington limit is already imposed in that part)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -48,6 +48,5 @@ jobs:
       run: |
         bash ci/compile_benchmark.sh
         pytest --cov=./cosmic --cov-report=xml
-        cd docs && make html; cd ../
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1
--- a/cosmic/src/evolv2.f
+++ b/cosmic/src/evolv2.f
@@ -2747,19 +2747,9 @@ component.
                endif
 *
             endif
-         elseif(kstar(j2).ge.10)then
-*
-* Impose the Eddington limit.
-*
-            dm2 = MIN(dm1,dme)
-            if(dm2.lt.dm1) supedd = .true.
-* Can add pulsar propeller evolution here if need be. PK.
-*
-*
-         else
+         elseif(kstar(j2).eq.3.or.kstar(j2).eq.5.or.kstar(j2).eq.6)then
 * We have a giant w/ kstar(j2) = 3,5,6
 *
-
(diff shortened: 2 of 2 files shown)


katiebreivik (commented):
In favor of clarity over compactness, I put in the overall accretion limit into each case. Now, for cases where an Eddington limit or nova outburst applies, we take the minimum of (acc_lim*dm1, dme) if acc_lim is >0. I think this is the clearest implementation of the flag for the user, but I'm open to conversations!

katiebreivik:
Alright @michaelzevin and @scottcoughlin2014: I've got the logic in the way I think it should be and I pulled the command to build the docs in the testing from the github actions since we build our docs separately anyway.

As a side note: do we want to keep the codecov/patch action going since it is still showing the checks as failing? I think we will never hit the target of 90% since we don't have a way to test fortran right now...

PR #493, opened 2021-03-19
rethinking convergence and filtering

### LINKED ISSUE #484: We need to allow a choose your own adventure filtering on what types of systems/evolutions you want to keep

### LINKED ISSUE #487: Issues with generating black holes with cosmicpop at near-Solar metallicity
There are a couple issues I've had in some runs, I'll try to list them logically below. The population run directory is on Quest at `/projects/b1095/michaelzevin/cosmic/populations/acc_efficiency/acclim_1/0.017`

1. When I ran this population using the multidim sampler, it wasn't finding _any_ black holes. This seems odd to me, but could _maybe_ be due to the initial conditions? So I ran with the independent sampler using `q_min=0.1`. However, I hope this will instead get fixed from addressing the issues below, since in versions of COSMIC from last summer I was still getting BBHs at Solar metallicity. 

2. When rerunning using the independent sampler, the model generated 54 matching BBH systems on the first step (which I saw in the log). However, it did not acquire **any** more systems in the rest of the run, which had a wall time of 5 days. I've attached the `.dat` file with these 54 systems. 

3. When the run hit its wall time of 5 days, rather than writing this in the `.out` file, it seems to have overwritten the `log` and `.out` files completely (i.e., they're empty). 

[dat_kstar1_14_kstar2_14_SFstart_13700.0_SFduration_0.0_metallicity_0.017.h5.zip](https://github.com/COSMIC-PopSynth/COSMIC/files/6131099/dat_kstar1_14_kstar2_14_SFstart_13700.0_SFduration_0.0_metallicity_0.017.h5.zip)

katiebreivik:
Renamed filter_bpp_bcm to filter_bin_state, reordered filtering so that we don't lose intermediate evolutionary states (e.g. xrbs) with the bin_state filtering.

Now the process is: 
(-) get the converging data set and bin_nums for conv_limits
(-) filter the converging data set on bin_state choice
(-) filter bcm/bpp/initC/kick to only contain data within conv_limits if user specified request to filter them, otherwise, don't
(-) compute convergence on conv dataframe with conv_limits

### DIFF
--- a/bin/cosmic-pop
+++ b/bin/cosmic-pop
@@ -347,8 +347,6 @@ if __name__ == '__main__':
         # Keep track of the index
         idx = int(bcm.bin_num.max()+1)
 
-        import pdb
-        pdb.set_trace()
         # If dtp is not set, filter out first timestep in bcm
         if np.all(dtp == IBT['tphysf'].values):
             bcm = bcm.loc[bcm['tphys'].isin(dtp)]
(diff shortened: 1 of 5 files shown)


scottcoughlin2014 (approved):
This looks good to me now, love the name change to those parameters.

PR #494, opened 2021-03-30
minor adjustments to acc_lim and don_lim flags

michaelzevin:
adjusted the acc_lim  and don_lim flags so that  -1 is the default BS…E, and for the acc_lim flag >=0 sets the amount of donor material accreted. This allows for acc_lim=0 to mean 'none of the donor material gets accreted'


katiebreivik (commented):
as long as the tests pass, this looks good to me!

scottcoughlin2014 (approved):
ya looks good makes sense to change this flags behavior.

PR #505, opened 2021-04-09
changes to the independent sampler

carlrodriguez:
Somewhat large commit that still needs some testing, but is pretty critical for CMC.  Two major changes to the independent sampler:

1. Adds a new flag to the independent sampler,  `sample_porb_first`, which if true will sample the orbital periods *before* the eccentricities.  This is somewhat necessary to prevent pathologically large eccentricities (e=0.999) that will come from the thermal distribution, and often place stars inside one another.  This breaks CMC, but is also unphysical for isolated binaries.  If this option is selected, the radii for Roche Lobe overflow are still computed with the orbits, but they are returned and used to truncate the eccentricity distribution (if it is not selected, then the eccentricities are sampled first, and the orbital periods are selected s.t. none of them are in Roche Lobe overflow)
2. I have changed it to where the minimum distribution for all the period distributions is set by the RL overflow criterion, not just the log-uniform option.  For the Sana12 distribution, it takes the maximum of the RL overflow criterion and the minimum of the actual distribution from Sana.

### DIFF
--- a/cosmic/sample/sampler/cmc.py
+++ b/cosmic/sample/sampler/cmc.py
@@ -42,6 +42,25 @@ def get_cmc_sampler(
 
     Parameters
     ----------
+    cluster_profile : `str`
+        Model to use for the cluster profile (i.e. sampling of the placement of objects in the cluster and their velocity within the cluster)
+        Options include:
+        'plummer' : Standard Plummer sphere.
+            Additional parameters:
+            'r_max' : `float`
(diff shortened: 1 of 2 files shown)


carlrodriguez:
Oh and third big change: I added another option, set_radii_with_BSE, that when True will use BSE to compute the radii for  the Roche Lobe in the independent sampler.  This is basically the same option that we were using in the IC generators for CMC, but I figured it was smarter to port it over to the independent sampler.

PR #521, opened 2021-05-12
Critical fix reguarding virial radii

carlrodriguez:
Changes how we specify the virial radius as a parameter; before we were NOT doing this consistently, and the orbital periods for binaries were wrong (the porb_max was always set assuming rvir=1, reguardless of what was set later).

Now, the virial radius and tidal radius are optional parameters to InitialCMCTable.sampler.  They can be reset when writing the hdf5 files, but that's optional


katiebreivik (commented):
I think this looks fine, but it looks like the test_king_profile is failing. Since new test data doesn't seem to be added, and the elson profile test doesn't fail, it might be worth another look from @carlrodriguez before we merge this in. (I don't know enough about cluster profiles to know why the tests would be failing for king but not elson)

carlrodriguez:
Turns out there was a bug in the old unit tests: the King profile test never actually generated a King profile, it just generated an Elson profile with the default parameters (which is what the unit test was).

PR #525, opened 2021-05-14
fixing NaN that appear for second omega angle

michaelzevin:
This small change fixes a (not-so) edge case. When both SNe have kicks of 0.0 (which is common for BBHs), the x_tilt that is calculated for the second SN is 0 since the angle between the angular momentum vectors is 0. When omega2 is then calculated, we get a divide by zero in a trigonometric function that leads to the NaN value in kick_info. 

I just added a simple if/else to address this edge case (setting omega2 to 0). I checked that [multiple binaries example on the Docs](https://cosmic-popsynth.github.io/COSMIC/examples/index.html#multiple-binaries) and it seems to have fixed things.

### DIFF
--- a/cosmic/src/kick.f
+++ b/cosmic/src/kick.f
@@ -417,7 +417,12 @@
      &               - comega*comega1*smu*smu1
 
         kick_info(sn,15) = ACOS(z_tilt)*180/pi
-        kick_info(sn,16) = ATAN(y_tilt/x_tilt)*180/pi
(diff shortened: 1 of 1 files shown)


scottcoughlin2014 (approved):
looks good and is just good not to have NaN's

katiebreivik:
agreed! This looks good to me! I'll go ahead and merge it in.

katiebreivik (approved):
👍

PR #529, opened 2021-05-18
Critical Fix to virial radii

carlrodriguez:
need to switch kwargs.pop to kwargs.get, since the virial radius was used in multiple places.


katiebreivik:
looks good!

PR #532, opened 2021-05-18
didnt quite check for the attributes properly with qmin and m2_min

katiebreivik:
This should finally do it!


michaelzevin (changes requested):
One quick change to the PR that I think would be useful: maybe first check that both q_min and m2_min are not specified. If they both are, we can fall back to using q_min, but we should spit out a warning.

katiebreivik:
Changes made @michaelzevin!

michaelzevin (approved):

PR #536, opened 2021-05-20
Better reff

carlrodriguez:
Ok, I've managed to speed up the radii computation by 60x by exporting it into BSE, rather than using COSMIC's sampling for it.  As a side note, this also eliminates the need to pass a params file or BSEDict when generating initial conditions


scottcoughlin2014:
This looks good!

Issue #432, opened 2020-08-27
Non consistent fallback fraction for delayed core-collapse prescription 


JRGarza:
At line 635 on the hrdiag.f module, the delayed core-collapse prescription, i.e. `remnantflag = 4`, rewrites the fallback fraction to a zero value, as `fallback = 0.d0`. This seems to not be consistent with Fryer+12, i.e. `fallback = 0.2d0 / (mt - mcx)`. As there is no other reference on that part of the code I wonder if this is a mistake.

https://github.com/COSMIC-PopSynth/COSMIC/blob/ca1b352ab9d062c851301bb98f2f8c587e5841c4/cosmic/src/hrdiag.f#L635

@michaelzevin @ssbvr

michaelzevin:
Hmmm...it looks like this is the case also at L1009. Both instances shouldn't affect anything, though, since `mt` is calculated correctly. 

I wonder if this change was made due to an update to Fryer's delayed prescription. @carlrodriguez you don't have any reason to think this would be the case, do you?

michaelzevin:
This is fixed in #537. We did digging and couldn't find the reasoning for why `fallback = 0.d0` was set, so reverted to what is done in Fryer+2012 exactly.

PR #537, opened 2021-05-21
ecsn fixes to fallback prescriptions

michaelzevin:
Addressing #432 

There seemed to be some adjustments to the Fryer fallback models for low core mass stars. I've changed so that it now follows Fryer+2012 directly, and also properly accounts for ECSN (including checking ecsn_mlow, which it wasn't doing before).

### DIFF
--- a/cosmic/src/hrdiag.f
+++ b/cosmic/src/hrdiag.f
@@ -533,7 +533,8 @@ C      if(mt0.gt.100.d0) mt = 100.d0
 * Use NS/BH masses given by Belczynski+08. PK.
 *
                      !First calculate the proto-core mass
-                     if(ecsn.gt.0.d0.and.mcbagb.le.ecsn)then
+                     if(ecsn.gt.0.d0.and.mcbagb.le.ecsn.and.
+     &                    mcbagb.ge.ecsn_mlow)then
                         mcx = 1.38d0
                      elseif(ecsn.eq.0.d0.and.mcbagb.le.2.25d0)then !this should be ecsn, unless ecsn=0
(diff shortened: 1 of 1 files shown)


cosmic/src/hrdiag.f
```diff
@@ -912,7 +913,8 @@ SUBROUTINE hrdiag(mass,aj,mt,tm,tn,tscls,lums,GB,zpars,
 * Use NS/BH masses given by Belczynski+08. PK.
 *
                      !First calculate the proto-core mass
-                     if(ecsn.gt.0.d0.and.mc.le.ecsn)then
+                     if(ecsn.gt.0.d0.and.mc.le.ecsn.and.
+     &                    mc.ge.ecsn_mlow)then
                         mcx = 1.38d0
                      elseif(ecsn.eq.0.d0.and.mc.le.2.25d0)then !this should be ecsn, unless ecsn=0
```
katiebreivik, line 919: We pulled lines 919-920 out in an earlier code block -- is there a reason that this block keeps them in?

katiebreivik (commented):
Looks like everything is good except the one comment I left. This is a shining example though that our evolve tests don't check for ecsn systems!

cosmic/src/hrdiag.f
michaelzevin: @katiebreivik I forgot to take these out for the Belczynski prescription remnantflag=2 (both in the normal and stripped-star cases). I've removed these in the most recent commit.

katiebreivik (approved):
Looks good! I'll go ahead and merge this in!

PR #540, opened 2021-05-25
Imf fix

carlrodriguez:
The kwargs for the custom IMF weren't actually being passed from CMC.  I've updated the code to do that, and also clarified the documentation.


katiebreivik:
thanks for catching this carl! I'll merge in now that the tests are passing.

PR #541, opened 2021-05-27
flow change to evolv2

carlrodriguez:
Ok, this is a tiny but significant change.  Basically when tphysf is sufficiently large (e.g. in CMC) and the CMC timestep is also sufficiently large (e.g. large cluster virial radius) you can run into a condition where COSMIC merges stellar binaries due to a combination of mass transfer/GW by subtracting angular momentum from the orbit (line 3281 in evolv2.f). 

However, if a binary is brought to 0 separation after that line (or ~1e-10), and the tphys if after the current tphysf, this binary can be returned to CMC having not been destroyed (even though the stars are well past contact), because the check for contact/collisions occurs *after* the goto at line 3557.  These binaries have ~10000x the binding energy of the whole cluster, causing CMC to crash when it checks for energy conservation.

To solve this, I've added a check in line 3281 of evolv2, whereby if the change in orbital angular momentum is greater than the total angular momentum, the goto at line 3557 is skipped and BSE checks for collisions before returning the binary to CMC.  

I've tested that this fixes the issues in CMC (it does) and doesn't change anything in COSMIC (at least for single binaries, it doesn't seem to).  However, I haven't run a full COSMIC population.

### DIFF
--- a/cosmic/src/evolv2.f
+++ b/cosmic/src/evolv2.f
@@ -204,7 +204,7 @@
       REAL*8 zero,ngtv,ngtv2,mt2,rrl1,rrl2,mcx,teff1,teff2
       REAL*8 mass1i,mass2i,tbi,ecci
       LOGICAL coel,com,prec,inttry,change,snova,sgl
-      LOGICAL supedd,novae,disk
+      LOGICAL supedd,novae,disk,inspiral
       LOGICAL iplot,isave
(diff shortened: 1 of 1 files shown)


katiebreivik:
Verified that this produces the exact same bpp array for populations of 10,000 DWD, BNS, and BBH progenitors with the PR and current develop branch so we are good to go!

PR #545, opened 2021-08-09
correcting the upper end of the pisn gap in the Marchant+ prescription

michaelzevin:
See Table 1 of [Marchant+2019](https://iopscience.iop.org/article/10.3847/1538-4357/ab3426/pdf). We were accidentally using the column with CO core mass rather than the column with He core mass for defining the top of the PISN mass gap in the Marchant prescription.


katiebreivik (approved):
good catch!

PR #546, opened 2021-09-01
Eddfac application for wind accretion in massive stars

katiebreivik:
Camille Liotine noticed some anomalously enormous X-ray emission coming from some BH binaries that were arising from super-Eddington accretion rates. We had a closer look and realized that while the Eddington limit is considered carefully and correctly for Roche overflow stable mass transfer, but not for winds. Normally, you'd not expect winds to produce super-Eddington accretion rates, but in our treatment for LBV-like wind mass loss we set the mass loss rate to 1.5e-4 MSun/yr. Standard Bondi-Hoyle accretion assumptions lead to very super-Eddington accretion unless the wind accretion is also limited.

This PR adds new code to limit wind accretion to the Eddington limit specified by eddfac both when the stars are detached as well as during Roche overflow mass transfer. In the event that wind mass accretion is already super-Eddington during Roche overflow, the mass coming through the Roche potential is not accreted at all. Otherwise, the Eddington limit is applied such that the Roche overflow mass is limited by the Eddington limit minus the amount of mass accreted from winds.

### DIFF
--- a/cosmic/src/evolv2.f
+++ b/cosmic/src/evolv2.f
@@ -2816,7 +2816,13 @@ component.
 *
          if(kstar(j2).ge.10)then
             if(acc_lim.lt.0.d0)then
-               dm2 = MIN(dm1,dme)
+*
+* If there is wind accretion the total amount of mass change is
+* dms(j2) = dmr(j2) - dmt(j2), where dmt(j2) is the accretion
+* from the companion. We should limit to the Eddington limit minus
+* the amount of accretion that is already coming in from Winds
+*
+               dm2 = MIN(dm1,dme + dms(j2))
 *
 * If we already hit supereddington wind accretion, don't add
 * any more mass through RLO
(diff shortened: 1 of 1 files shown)


cosmic/src/evolv2.f
```diff
@@ -2791,12 +2816,35 @@ SUBROUTINE evolv2(kstar,mass,tb,ecc,z,tphysf,
...
                if(dm2.lt.dm1) supedd = .true.
             elseif(acc_lim.ge.0.d0)then
-               dm2 = MIN(acc_lim*dm1,dme)
+*
+* If there is wind accretion the total amount of mass change is
+* dms(j2) = dmr(j2) - dmt(j2), where dmt(j2) is the accretion
+* from the companion. We should limit to the Eddington limit minus
+* the amount of accretion that is already coming in from Winds
+*
+               dm2 = MIN(acc_lim*dm1,dme + dms(j2))
```
michaelzevin, line 2839: shouldn't this be `dme  - dms(j2)`? Or is `dms(j2)` negative?

michaelzevin (changes requested):
Looks pretty good! Just one question below.

cosmic/src/evolv2.f
katiebreivik: Since dms(j2) = dmr(j2) - dmt(j2), dms(j2) is negative for CO accretors. This especially true since dmr(j2) << dmt(j2) because the wind mass loss is 0.

michaelzevin (approved):

PR #549, opened 2021-10-06
Fixing the final kstar range implement

chiragchawla134:


katiebreivik (approved):
looks good! thanks @chiragchawla134!

katiebreivik:
ack! Looks like we broke the tests (which include the same bug that you fixed @chiragchawla134) I'll submit a quick test pr to fix this, then you can merge that, then we can merge. Apologies!

PR #550, opened 2021-10-25
fixed the silly integration errors when sampling a Plummer profile

carlrodriguez:
Thought I already fixed this, but the PR from a few months ago was empty and the changes were actually sitting on my desktop.  If gamma=4, the Elson profile will generate velocity samples directly from the Plummer distribution function (instead of numerically integrating the distribution function like it does for other cases)


katiebreivik (approved):
looks good to me. thanks for catching this!

PR #551, opened 2021-10-26
qmin, m2min, msort

michaelzevin:
reworked secondary and binary sampling to allow for both qmin and m2min to be set, as well as reworked the workflow when msort is specified.


katiebreivik (approved):
this is great! thanks @michaelzevin !

PR #555, opened 2022-04-11
Sse changes

poojanagrawal:
This pull request is for the following changes in mlwind.f.
1. I have added `zsun_wind` to differentiate it from the global `zsun` variable and have set its value to 0.019  as wind mass-loss rates have originally been scaled from this value (e.g., see Section 4 of Vink et al. 2001). 
2. As described in Section 8 of Vink et al. 2001, the ratio of terminal velocity to escape velocity is dependent on metallicity by a factor of 0.13. Thus, I have added the appropriate scaling factor. 
3. Following the Dutch wind prescription of MESA, I have extended the use of Vink winds beyond 50 000K.  This is only relevant for hot, very massive stars (M>150Msun), that have not yet reached the Eddington limit. Before this, winds from Nieuwenhuijzen & de Jager 1990 were getting used for such stars.

### DIFF
--- a/cosmic/src/mlwind.f
+++ b/cosmic/src/mlwind.f
@@ -5,7 +5,7 @@
       integer kw,testflag
       real*8 lum,r,mt,mc,rl,z,teff,alpha
       real*8 dml,dms,dmt,p0,x,mew,lum0,kap
-      real*8 MLalpha, zsun_wind
+      real*8 MLalpha, zsun_wind, V
       external MLalpha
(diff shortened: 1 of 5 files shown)


katiebreivik:
Thanks for opening this @poojanagrawal! I remember from our chat that we'll need to update some of the unit tests. I'll have a careful look while the tests run.

katiebreivik:
Alrighty this is good to go!

PR #563, opened 2022-08-08
Orbital sampling

elenagonzalez870:
Some binaries were being sampled at separations smaller than RL radius, causing them to merge immediately. To fix this, I added lines in the independent.py file that check if the orbital period sampled is less than the minimum orbital period for RL. If so, it replaces the value with the minimum RL separation. The same procedure is used for the sana12 and renzo19 samplers. For consistency, I also truncate the eccentricities to avoid RL overflow. For the sana12 eccentricity sampling, I set the maximum value to be either the sana12 value of 0.9 or the one corrected for RL, whichever is smaller.


katiebreivik:
Thanks @elenagonzalez870 !

PR #567, opened 2022-08-11
updating versioneer

katiebreivik:
- removed use2to3
- updated versioneer


katiebreivik:
I think this isn't necessary right now actually!

PR #568, opened 2022-08-11
switching out pypi publish workflow

katiebreivik:
moving to [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) for wheel building


katiebreivik:
this is super minor so no review!

PR #570, opened 2022-09-01
Implementing Eddington limit for BHs based on spinup/ISCO

michaelzevin:
This PR adjusts the Eddington limit for BHs based on the implementation in [Marchant et al. 2017](https://ui.adsabs.harvard.edu/abs/2017A%26A...604A..55M/abstract), following Equations 3-5. It introduces a new global parameter `Mbh_initial` that remembers the initial mass of the BH at formation, and calculates the `eta` parameter that alters the mass transfer rate based on where the ISCO is, assuming the BH is spinning up from stable accretion as in [Thorne 1970](https://ui.adsabs.harvard.edu/abs/1974ApJ...191..507T/abstract). 

As of now, I added an additional global parameter `BHbirth_ind` that is 0 before the BH forms and 1 after so that `Mbh_initial` doesn't get overwritten in `hrdiag.f`, though there's probably a more glamorous way to ensure this. 

This actually causes the Eddington limit for BHs to be ~1 order of magnitude higher for low-spinning BHs than the standard BSE prescription!!! It is unclear where Hurley got the expression for the Eddington limit that is implemented in stock BSE (Equation 67 of [Hurley et al. 2002](https://ui.adsabs.harvard.edu/abs/2002MNRAS.329..897H/abstract)). The two expressions are identical for `eta =  0.5`, which is slightly above the maximum value of `eta=0.42` which is attained for a maximally-spinning BH (one that accreted more than sqrt(6) times its initial mass). 

Attached are the mass transfer rates for a few binaries, comparing the current BH accretion limit to the one in this PR. There is still an issue of deltam_1 for the first timestep written into the bcm array not obeying the imposed Eddington limit, which can be seen in a few of these systems. 

eddlim

There are probably other changes that need to be made to the unit test files, so let me know what else needs to be done!

### DIFF
--- a/cosmic/src/hrdiag.f
+++ b/cosmic/src/hrdiag.f
@@ -803,7 +803,6 @@ C      if(mt0.gt.100.d0) mt = 100.d0
 * Store the initial BH mass for calculating the ISCO later
                      if(BHbirth_ind.eq.0)then
                         Mbh_initial = mt
-                        WRITE(*,*) Mbh_initial
                         BHbirth_ind = 1
                      endif
                   endif
@@ -1185,7 +1184,6 @@ C      if(mt0.gt.100.d0) mt = 100.d0
 * Store the initial BH mass for calculating the ISCO later
                      if(BHbirth_ind.eq.0)then
                         Mbh_initial = mt
-                        WRITE(*,*) Mbh_initial
                         BHbirth_ind = 1
                      endif
                      endif
@@ -1285,7 +1283,6 @@ C      if(mt0.gt.100.d0) mt = 100.d0
 * Store the initial BH mass for calculating the ISCO later
          if(BHbirth_ind.eq.0)then
             Mbh_initial = mt
-            WRITE(*,*) Mbh_initial
             BHbirth_ind = 1
          endif
          lum = 1.0d-10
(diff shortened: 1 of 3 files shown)


cosmic/src/hrdiag.f
```diff
@@ -800,6 +800,11 @@ SUBROUTINE hrdiag(mass,aj,mt,tm,tn,tscls,lums,GB,zpars,
                         endif
                      endif
                      mt = mrem
+* Store the initial BH mass for calculating the ISCO later
```
katiebreivik, line 803: What do you think about initializing Mbh_initial to 0, then checking if Mbh_initial.eq.0 instead of BHbirth_ind?

katiebreivik (commented):
I think my question above isn't 100% necessary, but will save on new variable bloat if it works. Otherwise, this looks good to go!

cosmic/src/hrdiag.f
michaelzevin: Yep makes total sense and is definitely cleaner, I’ll push that change to this PR.

michaelzevin:
@katiebreivik I've done what you suggested. Looks like the unit tests never encountered super-Eddington accretion so they still pass. 

There's still the issue with the deltam1_bcm for the first timestep of the BCM array sometimes still being super-Eddington, though I don't think this is a huge issue.

katiebreivik:
Hey @michaelzevin sorry for being slow! -- can you open an issue with an example binary that still goes super Eddington on the first time step? I just want to make sure we don't lose it in the mix. Once that issue's open, feel free to merge!

michaelzevin:
Good call. I've opened an Issue at #571. I'll now go ahead and merge.

PR #574, opened 2022-10-09
Exchange some `print` statements for `warnings.warn`

TomWagg:
Hi! [I've been using](https://github.com/TomWagg/cosmic-gala) `COSMIC` with @katiebreivik and love it 😁 One small thing that was bugging me was the following message I got a lot
> 'zsun' is set to a different value than assumed in the mlwind prescriptions (you set it to '0.014' and in mlwind, zsun_wind=0.019)

So I just made a small PR to convert any print statements to `warnings.warn`. This shouldn't change anything but it does now allow me to use `warnings.filterwarnings` to ignore certain messages like this one about the zsun value 🙂 

I also wasn't sure on the changelog/version updating etiquette but I changed those too.


katiebreivik:
Thanks for doing this @TomWagg  -- I've been meaning to figure this out for a while, so this is very much appreciated! Thanks also for updating the changelog and versions.

katiebreivik:
Yay! Tests passing -- merging now.

PR #582, opened 2023-05-02
Hardcoded zsun for winds plus other bug fixes following Gaia BH project

poojanagrawal:
This PR includes modifications/bug fixes in cosmic from the Gaia BH project.

1. To avoid accretion of matter from a massless companion, the calculation of mass changes in evolv2.f now excludes type 15 stars.  The error results from non-zero 'dme' and the fact that the minimum of NaN (which is 'dmt') and dme gives me, leading to non-zero 'dmt' .

2. The properties of massless remnants are now exclusively assigned in hrdiag.f, so that they do not interfere with the evolution of other stars (e.g. due to non-neglible radii).

3. Added a new evolve_type (100) to identify systems that fail to evolve due to iter>loop error. 

4. Following the use of qcrit, replaced kstar(k) to kstar(k)+1 by PA to avoid errors for type 0 stars.

5. Added an extra call to write bpp array for systems that become remnant on entering evolv2.f ( mostly systems whose evolution has been restarted), and avoid write bpp array altogether, leading to issues later in the code. 

6. (Note that Gaia BH project did not have this.) Wind calculation should use original zsun values from their respective wind prescription and not depend on the zsun elsewhere in the code. The value is hardcoded to avoid misuse of zsun in mlwind.f.

### DIFF
--- a/cosmic/sample/initialcmctable.py
+++ b/cosmic/sample/initialcmctable.py
@@ -308,8 +308,8 @@ class InitialCMCTable(pd.DataFrame):
         singles_bottom = pd.DataFrame(
             np.zeros((1, Singles.shape[1])), index=[0], columns=Singles.columns
         )
-        singles = singles.append(Singles)
-        singles = singles.append(singles_bottom)
+        singles = pd.concat([singles, Singles])
+        singles = pd.concat([singles, singles_bottom])
         singles["r"].iloc[-1] = 1e40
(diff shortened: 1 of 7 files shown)


katiebreivik (commented):
This is great -- we still need to add in string conversions for evolve_type == 100.0

katiebreivik (approved):
👍

PR #592, opened 2023-07-24
RTMSFLAG

kulvi90:
Added new flag (rtmsflag) to take the metallicity effects into account while calculating the stellar radius at the main-sequence turnoff (rams).  All the relevant functions for the calculations are in zfuncs.f file.

rtmsflag = 0 ->  uses the sse rtms for M < 200M_sun and extrapolation for z < 0.0008 and M>200m_sun.
rtmsflag = 1 -> interpolates the rtms from Boost tracks (Szécsi et al. (2022)). Extrapolation is used after M > 575 M_sun.
rtmsflag = 2 -> uses the best-fit power law for rtms vs stellar mass from BPASSv2.2 tracks (Stanway & Eldridge (2018)).  rtms data for BPASS tracks is available only till 300 m_sun and beyond that we follow the fitted power law profile.


katiebreivik:
Thanks @kulvi90! This is great!

PR #593, opened 2023-08-01
Quick fix for deprecation due to incompatibility with Pandas >= 2.0.0 updated API

smandhai:
df.append has become deprecated in pandas versions >2.0.0 (latest == 2.0.3). This was preventing COSMIC from stacking tables and would cause it to return errors if the local pandas installation was incompatible (newer). Other .append (lines 382, 438-442,) methods will need to be replaced or worked around in a similar fashion to ensure compatibility going forward with future versions of pandas.

### DIFF
--- a/bin/cosmic-pop
+++ b/bin/cosmic-pop
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/home/u77526sm/anaconda3/envs/cosmic_test/bin/python3.10
 
 # Code: cosmic-pop.py
 # Version: 1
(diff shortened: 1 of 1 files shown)


katiebreivik:
Thanks @smandhai -- I have definitely been meaning to fix this and had not gotten around to it. If you prefer, I can do a quick search and replace if you prefer to clean up the rest of the appends. Let me know what you think-

smandhai:
I'm more than happy to help implement the workaround to all the appends. I'm hesitant about removing them altogether, but I've tested my fix and it seemingly works with COSMIC 3.4.7.  I'll get this worked into my forked version

smandhai:
Now done - but I think the exceptional cases will need to be tested properly - especially the NaN case.

katiebreivik:
Hey @smandhai -- I finally got around to testing this on a fairly large (1e6) population and everything seems to check out so I'm going to go ahead and merge. 

Thanks for your help!

smandhai:
Hey @katiebreivik , that's good news and happy to help! I've noticed another issue/bug with the latest version of pandas. I ran 3/4 COSMIC batches which worked fine for a billion systems but the 4th triggered an issue with the pd.concat routines. There were missing brackets (odd these didn't flare up earlier). I've now fixed these, it should be in the latest commit (for comparison). Apologies for not catching this earlier!

katiebreivik:
Ah -- good find @smandhai!

Can you re-open this PR and I can re-merge?

smandhai:
I've had to submit a new pull request - https://github.com/COSMIC-PopSynth/COSMIC/pull/603 

This should resolve that issue but I haven't been able to test it on scale yet!

PR #597, opened 2023-08-23
Removed the conditional for MESA RTMS (RTMSFLAG = 3) because the func…

kulvi90:
Removed the conditional for MESA RTMS (RTMSFLAG = 3) because the function

rtmsMESA to calculate rtms for given mzams was not implemented for MIST models. It would have given an error if RTMSFLAG = 3 were used.

Added a few important comments regarding the specific metallicities implemented for BoOST and BPASS. NOTE: Only a few common metallicities (Z <= 0.002) between BoOST and BPASS are implemented.

Included rtmsflag in files test/test_utils.py and utils.py


katiebreivik (approved):
Thanks @kulvi90!

PR #599, opened 2023-08-28
Corrected the metallicity threshold for the cosmic conditional

kulvi90:
Changed the metallicity threshold in rtmsf function of zfuncs.f for the cosmic conditional from 0.01 to 0.0008.

Cleaned up the code a bit and added few lines of comments in zfuncs.f


katiebreivik (approved):
👍

PR #602, opened 2023-09-27
Allow user to specify a target total mass for independent sampler

TomWagg:
Hi! I mentioned to @katiebreivik that I was planning to make a sampler that targeted mass instead of size and she suggested I make a PR so here I am.

**Big picture changes I made:**
- Added `total_mass` parameter, which (in conjunction with `sampling_target = 'total_mass'`) makes it so that the sampler keeps going until the total sampled mass reaches a certain value (rather than the number of binaries reaches a certain value)
- Added a `trim_extra_samples` parameter which avoids overshooting the total mass and only retains the samples that get you as close as possible to the desired total mass (so it's almost always very close but as far as 2 x primary_max away)

**Some gritty details:**
- I changed some variable names because it was a bit confusing having `mass_binaries`, `mass1_binaries` _and_ `mass1_binary` 😅 
- I used a lambda function for evaluating the target to avoid checking an if statement for every sampling loop instead of just once
- I switched some `np.sum` to `sum` - I think these are much faster for lists (where opposite is of course true for np arrays)

**Testing**
Unsure exactly how testing is done with COSMIC but I figure I should add some simple unittests e.g.

1. Sample size always >= size for `sampling_target=size`
2. Sample mass always >= total_mass for `sampling_target=total_mass`
3. Sample mass - total_mass <= 150 for `sampling_target=total_mass`
I'll ask @katiebreivik about this tomorrow and add these to the PR 🙂


TomWagg:
**Separately a question:** The way the samples get masked seems overly complex to me but potentially missing something. But would it not work to turn this
```
        # select out the primaries and secondaries that will produce the final kstars
        (ind_select_primary,) = np.where(
            (mass1_binaries > primary_min) & (mass1_binaries < primary_max)
        )
        (ind_select_secondary,) = np.where(
            (mass2_binaries > secondary_min) & (mass2_binaries < secondary_max)
        )
        ind_select = list(
            set(ind_select_primary).intersection(ind_select_secondary))
```
into this?
```
        ind_select = ((mass1_binaries > primary_min)
                      & (mass1_binaries < primary_max)
                      & (mass2_binaries > secondary_min)
                      & (mass2_binaries < secondary_max))
```

TomWagg:
And finally here's a quick plot of how close I get to the target_mass when running 10000 samples, each targeting 15000 solar masses. Mostly pretty close but it all comes down to the IMF and if I get unlucky with a big binary right at the end
image

TomWagg:
@katiebreivik I've added the documentation that you suggested, let me know what you think! 🙂 (I also spruced things up a little with some extra heading, sphinx admonitions and api refs - hope that's okay!)

katiebreivik:
👍

PR #610, opened 2023-11-15
Timestep conditions: implement validation, add `teff` as an option

### LINKED ISSUE #609: [general issue] bcm `timestep_conditions` fail silently for missing columns
Currently if you do timestep conditions for a column that COSMIC doesn't allow (or just have a typo), the condition won't be applied but you also won't be told that something went wrong.

PR incoming!

TomWagg:
Two main sets of changes here:

- Validation: `checkstate.py` now checks whether a column is actually present in the CHECKSTATE_COLUMNS and throws an error if it isn't rather than fails silently. I also added a check that there aren't more than 15 columns if some insane users wants to condition on 16 variables because that will break it haha
- New condition variables: `teff_1` and `teff_2` can now be used in timestep_conditions just like the other variables 😄 

I also added some instructions to `checkstate.f` in case anyone else needs to add more condition variables one day.

p.s. This is my first attempt at coding in fortran so please do check I didn't break something (it doesn't _seem_ like I have but just in case).

This will fix #609


katiebreivik (approved):
Thanks @TomWagg  -- the fortran looks good to go!

PR #611, opened 2023-11-29
Plot binary evolution improvements

TomWagg:
Here are a couple of improvements/fixes for `plot_binary_evol`:

- Fix usage of gridspec to avoid double axes that overlap and confuse things (see screenshot of current docs page below)
- Add option to supply a `t_min` and `t_max`
- Add option to supply custom colours and labels for the ktypes

Here's a little test script to see how things work:
```
from cosmic.plotting import plot_binary_evol
import matplotlib.pyplot as plt

plot_binary_evol(bcm,
                 ktype_kwargs={'k_type_labels': ["TOM" for i in range(16)]},
                 t_min=100,
                 t_max=200)
plt.show()
```

Before:

image

After (different binary though):
image

### DIFF
--- a/cosmic/plotting.py
+++ b/cosmic/plotting.py
@@ -543,18 +543,17 @@ def plot_binary_evol(bcm, sys_obs={}):
         Figure containing all the plots!
     """
 
-    fig, ax = plt.subplots(5, 1, figsize=(10, 10))
+    fig = plt.figure(figsize=(10, 10), layout="tight")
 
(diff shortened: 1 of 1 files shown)


katiebreivik (approved):
Thanks @TomWagg  -- it's good to have t_min and t_max added into the plotting only function; I had thought we were good to go, but only used evolve_and_plot which uses the time limits.

PR #613, opened 2023-12-19
added the necessary updates for including a central_bh in CMC

carlrodriguez:
work from Tomás and me on including central massive black holes (MBHs) in CMC.

This places an MBH at the 0 index of the Singles array.  When using the regular CMC sampler the mass is specified in solar masses; when using the cmc-point-mass sampler the mass is specified in number of stars (e.g. for a 10,000 point-mass cluster, central_bh=1000 yields an MBH with 10% of the cluster mass).

There's an additional parameter, scale_with_central_bh, which can be passed to the sampler; if True, this adds the MBH potential energy into the calculation when rescaling the stellar positions (default is False).

finally, added a ``seed'' parameter to cmc_point_mass to create different point-mass realizations of clusters.


katiebreivik:
Thanks for adding in the changelog @carlrodriguez !

katiebreivik (approved):
👍

PR #616, opened 2024-02-12
Added a flag wd_mass_lim

claireshiye:
Added a flag wd_mass_lim controlling if massive remnants (>Mch) from MIC are allowed.

Fixed a bug in evolv2.f where in elseif(kstar(j1).eq.10.or.kstar(j2).gt.10)then the or should be and.

The effects of the flag are partially demonstrated in this paper https://ui.adsabs.harvard.edu/abs/2024ApJ...961...98Y/abstract.


katiebreivik:
👍

PR #620, opened 2024-04-11
Created maximum wall time option

xevra:
For some applications, it's useful to end COSMIC gracefully after a certain amount of time has passed.

On the Open Science Grid, for example, a job is limited to 20 hours. If the job would go beyond 20 hours, it would be killed and data would be lost. Due to the nature of match filtering, it's not always possible to predict how long a COSMIC simulation will take to run.

In my applications, I would like to end the simulation at 19 hours to give the data time to download back from the run node, so I can process the outputs. Ending the process from the shell or launcher is an option which would enable the recovery of data and log files, but may interrupt COSMIC at an inopportune time.

The maximum wall time option enables gracefully ending a COSMIC job after a set amount of time, so that COSMIC will always end at the end of a loop.

The default value of 3155760 seconds corresponds to a little over a month. While this does technically change the default behavior of COSMIC, it will only affect simulations which would otherwise be running for more than one month (wall time).

### DIFF
--- a/bin/cosmic-pop
+++ b/bin/cosmic-pop
@@ -98,6 +98,8 @@ def parse_commandline():
                         help="Number of binaries to try before checking for "
                         "convergence, it will check ever Nstep binaries until "
                         "it reach Niter binaries", type=int, default=10000)
+    parser.add_argument("--max-wall-time", type=int, default=3155760,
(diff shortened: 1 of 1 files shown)


katiebreivik (approved):
after many many days (sorry!): 👍

PR #628, opened 2024-06-04
Metisse integration meson

poojanagrawal:
(re-)added METISSE to meson supported verion of COSMIC.


katiebreivik:
thanks @poojanagrawal

PR #629, opened 2024-06-04
added METISSE as submodule

poojanagrawal:
do  `git submodule update --init --recursive` after pulling


katiebreivik:
merging for testing on local machines rather than actions for finding the submodule

PR #632, opened 2024-06-06
Stop statements

poojanagrawal:
When calling Fortran from Python, if Fortran code stops for some reason, the Python part doesn't stop by itself. Instead, it waits patiently for Fortran to return something (until the user's patience runs out or the code's runtime). With this pull request, METISSE now sends an error message back to Python when a Fatal Error occurs in zcnsts.f. For later stages of the code, METISSE continues to run with some dummy values but returns an error to set evolve_type to 101: METISSE error.


katiebreivik:
woohoo!

PR #644, opened 2024-06-21
offner22 primary mass dependent binary fraction

MarkGM02:
I updated the initial binary sampler, adding a new binary fraction model [offner22](https://arxiv.org/abs/2203.10066).


katiebreivik:
3.8 be damned. We'll merged it anyway! Thanks @MarkGM02!

PR #646, opened 2024-06-24
added calls to set star type as rejuvenated for very low mass stars

poojanagrawal:
METISSE treats the evolution of normal stars differently from the rejuvenated stars and a call to set_star_type is needed to reset the stellar type. Added the appropriate calls for the rejuvenation of very low mass stars.


katiebreivik:
thanks @poojanagrawal

katiebreivik (approved):
👍

PR #647, opened 2024-06-28
Add debugging setup, avoid NaNs from timestep issue

### LINKED ISSUE #608: [general issue] NaN results produced in `bpp` table (when evolve to HeHG during stable MT)
When running `COSMIC` on large populations of binaries I frequently run into edge cases that result in NaN values in the `bpp` output. I've been chatting with @katiebreivik about this and realised I should probably organise a little. So here's an issue about this and I'll keep a list of commonalities that seem to result in NaNs up to date and post `initC` and `bpp` tables in the comments below.

**NaN clues noticed so far:**

1. This appears to consistently happen when a star evolves on HeHG (kstar=8) _during_ mass transfer
2. Often the `RRLO_1` column is the first one that seems to get a NaN
3. I've only ever seen this happen with binaries (I ran simulations with binfrac=0.5 and kept all of the singles)

edit: GitHub won't let me upload hdf5 files so I'm going to attach separate CSVs instead!

Here's a quick confirmation that it is happening whenever you have a star evolving to HeHG during mass transfer:
```
nan_bpp = pd.read_csv("nans_bpp.csv", index_col=0)
nan_bin_nums = nan_bpp["bin_num"].unique()
hehg_during_mt = [False for _ in range(len(nan_bin_nums))]
for i, bin_num in enumerate(nan_bin_nums):
    single_bpp = nan_bpp.loc[bin_num][["kstar_1", "kstar_2", "evol_type", "RRLO_1", "RRLO_2"]]
    mt = False
    for j in range(len(single_bpp)):
        if single_bpp.iloc[j].isnull().any():
            if single_bpp.iloc[j]["kstar_1"] == 8 and mt:
                hehg_during_mt[i] = True
                break
        
        if single_bpp.iloc[j]["evol_type"] == 3:
            mt = True
        elif single_bpp.iloc[j]["evol_type"] == 3:
            mt = False
            
print(all(hehg_during_mt))

>>> True
```

TomWagg:
This PR does two main things:

1. After a several month long battle, this PR solves #608 with an offensively small number of changes to the code - the change avoids NaN values in particular evolution cases (see below)
2. It adds `debug` and `.vscode` folders so that COSMIC developers can quickly and easily debug the fortran code using vscode

**Explanation of the NaNs**
I was finding that NaNs were occurring during complicated parts of evolution (e.g. He star stellar type changing during stable mass transfer). The reason for this was that there was a part of the code that reverted to using large timesteps if the number of iterations exceeded 1000. This means the star could suddenly jump forward 5 Myr while it was transferring mass...and then everything blows up. As a simple fix for now I've changed 1000 to `loop`, which is set to 20,000 by default - this makes it more consistent and also clearly increases the limit a lot. This fix removes every case of NaNs that I have in my saved files.

It's important to keep in mind for future reference though **this may not be solving the underlying issue**. Large timesteps messing with things seems to imply there's some logic that is violating an order-of-operations situation and causing NaNs but we can skirt around it with small enough timesteps. Something to keep in mind if NaNs arise in future 🤷🏻 

**Debugging details**
The `.vscode` folder adds scripts that automatically compile and execute a testing script which takes a `binary.in` file as input. I've included a python script to auto create these input files from `initC` rows. When I do the docs PR I'll include a proper tutorial for developers to get this up and running :)


katiebreivik:
yay! we love to get rid of NaNs!

PR #652, opened 2024-08-07
New ``gamma`` option for circumbinary discs

TomWagg:
Added a new ``gamma`` prescription following Zapartas+17 Eq. 9 (and originally Artymowicz & Lubow (1994)). This assumes assumes mass is lost through the outer Lagrangian point, forming a circumbinary disk.


katiebreivik (approved):
👍

PR #659, opened 2024-08-09
Adding feature to extract commit hash

elenagonzalez870:
With this addition, when cosmic is installed, the file `_commit_hash.py` is created in `src/cosmic`. 

If the user imports cosmic and types `commit.__commithash__`, this will return the commit hash from this version of COSMIC.


katiebreivik:
thanks @elenagonzalez870 !!

PR #661, opened 2024-08-13
Allow a gamma of -3 in ``error_check``

TomWagg:
I missed the check for different allowed ``gamma`` values sorry!


katiebreivik:
👍

PR #673, opened 2024-09-08
Allow user customisation of columns in `bpp` and `bcm` tables

TomWagg:
I've added two arguments to `Evolve.evolve()`, `bpp_columns` and `bcm_columns`. Users can specify a list of columns to include in either table, by default it will use the current setup so no breaking changes with this 😅 I've made the changes directly in the Fortran as well so this will perhaps cut down on RAM usage during the run.

The only change I made that alters the default behaviour is to report `kstar_x` and `evol_type` as integers instead of floats, which I think makes sense?

(Selfishly motivated by TomWagg/cogsworth#86 🙃)

### Implementation details

The way it works is:

- Define a list of all possible columns (in python and in fortran files)
- Convert user input into a number of columns in each table and indices of the variables you want to add
- In `writetab` (new function that combines `writebpp` and `writebcm` in a simpler way), access the specific variables in the order the inds specify

This means I had to define some new variables in `const_bse.h` and also change the calling of `writebpp` and `writebcm` everywhere.

### Demo

Default behaviour
image

Specifying certain columns
image

### DIFF
--- a/src/cosmic/evolve.py
+++ b/src/cosmic/evolve.py
@@ -360,8 +360,9 @@ class Evolve(object):
         col_inds_bpp[:len(bpp_columns)] = [ALL_COLUMNS.index(col) + 1 for col in bpp_columns]
 
         # save bpp column information in the initial conditions
-        initial_conditions[0]["n_col_bpp"] = len(bpp_columns)
-        initial_conditions[0]["col_inds_bpp"] = col_inds_bpp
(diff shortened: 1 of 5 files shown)


katiebreivik (approved):
amazing! We can merge this in for now but we should be sure to add in some docs for providing the kwargs for the columns to the evolve call.

PR #678, opened 2024-10-02
Implement new kick prescription

### LINKED ISSUE #675: Kick implementation peculiarity: why does secondary star ejection velocity depend on natal kick strength?
I was chatting to @katiebreivik about this and we thought it would be great to hear what @michaelzevin thinks about this :)

My intuition is that when a primary star receives a natal kick that disrupts the binary, the secondary star is ejected with close to its orbital velocity. However, it seems in COSMIC that the ejection velocity of the secondary is very dependent on the natal kick that the primary receives.

As an example, I ran the same 20+15 Msol binary (P=100d, ecc=0, Z=0.02) 5000 times to get it to sample different natal kicks and found the following distribution of ejection velocities.
image

The peak of the distribution lines up with the orbital velocity of the secondary but there's a large spread of different values.

Also if I re-run the same code but with extremely ridiculous kicks (`sigma=10000`) then the ejection velocities _also_ increase a lot

image

Does this behaviour make sense? My guess would have been that the ejection velocity isn't affected by the natal kick significantly.

TomWagg:
# tldr;

This PR replaces the default COSMIC prescription for handling kick velocities. Instead of using Kiel & Hurley 2009, it now follows Pfahl+2002. The former produces strange results for disrupted systems, in which the secondary ejection velocity showed a dependence on the natal kick strength, which goes against physical intuition (see #675).

# Changes

Breaking code changes:

- Kick info table is now (2, 18) not (2, 17), where the extra column is included so every Euler angle can be recorded
- `kickflag` is now an option that can be 1, 2, 3, or 4 to replace the old options. Additionally, if users want to use the old K&H prescription, you use the negative of any of those values.

Other code fixes (**note:** in the interest of reproducibility I haven't fixed these in the old Kiel prescription)
- A coordinate transformation for secondary stars reaching supernova in an already disrupted binary was wrong (assumed co-lateral coordinates erroneously), I've fixed this
- Not sure if this is a fix per-se but the criteria for collisions were weird in the old version, I now do proper vector calculations to find their closest separation after the SN and check if it is within the secondary radius

Physical evolution changes
- Ejection velocities for secondaries are now much more tightly distributed around the pre-SN orbital velocity

# Comparison of a disrupted system

For unbound systems, the velocity of the companion has been changed a fair bit. Here's 5000 random draws for the same binary (m1=20, m1=15, porb=200, Z=0.02, ecc=0)

image
image

# Comparison of a bound system

And for bound systems, the final separations are pretty much unchanged (these are just different random draws so you see a little scatter).

image
image

This will close #675.


katiebreivik (approved):
This looks great! if you can add a test to make sure that we have a way to test whether or not a future development for the kick subroutine breaks what you have now, then we are good!

PR #679, opened 2024-10-08
Fix NaNs coming from supernova kicks

TomWagg:
This fixes two different bugs that result in NaNs in the `bpp` and `kick_info` tables. Here's details on both:

# NaN in $\psi_{\rm Euler}$

This one's my fault 😅 In cases where **both** the natal kick is 0km/s _and_ the pre-SN eccentricity is 0.0 then one cannot compute the cross product (a) between the orbital angular momentum pre/post-SN or (b) between the pre/post-SN Laplace-Runge-Lenz vector. I've added conditions for this situation and just use a random angle in this case.

**Example system  for reproducing:** `["mass_1", "mass_2", "porb", "ecc", "metallicity"] = [51.824108, 24.551823, 10.904105, 0.071088, 0.001284]`

# NaN in `porb`

Based on guesswork alone, I think this one is coming from someone adding a call to `COMENV()` but not also copying the code below it which is necessary for bookkeeping. This means that if a supernova occurs **during a CE** the `porb` post-SN is set to a NaN because the `binstate` hasn't been updated. I added the same code that's below the other 2 calls to `COMENV()`. @katiebreivik you should check that this should all be here...I _think_ it should but something to look at closely.

**Example system for reproducing:** `["mass_1", "mass_2", "porb", "ecc", "metallicity"] = [46.211960, 15.483146, 6639.574181, 0.589395, 0.00192]]`

Slowly winning our campaign against NaNs 🎉

### DIFF
--- a/src/cosmic/src/evolv2.f
+++ b/src/cosmic/src/evolv2.f
@@ -2402,6 +2402,32 @@ component.
      &               jp,tphys,switchedCE,rad,tms,evolve_type,disrupt,
      &               lumin,B_0,bacc,tacc,epoch,menv,renv,bkick,
      &               deltam1_bcm,deltam2_bcm)
+         if(j1.eq.2.and.kcomp2.eq.13.and.kstar(j2).eq.15.and.
+     &      kstar(j1).eq.13)then !PK.
+* In CE the NS got switched around. Do same to formation.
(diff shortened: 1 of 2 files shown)


katiebreivik (approved):
finally approving @TomWagg

PR #688, opened 2025-01-08
Avoid NaNs from massless remnant kicks, fix debugging setup

### LINKED ISSUE #687: [general issue] PISN in common envelope borks kicks in Pfahl prescription
# Describe your issue in as much detail as possible -- if your issue arises from a cosmic-pop command line call, please include the exact syntax of the command line call in your description.

@xevra has found a lovely edge case in BBH formation where we get NaNs that cause the fortran to _**hang indefinitely**_ 😱

This ipython block with the attached Params file will cause the problem:

```from cosmic.sample.initialbinarytable import InitialBinaryTable
from cosmic.evolve import Evolve
import numpy as np
import pandas as pd
dat_in = pd.read_csv('some_binaries.txt', header=0, delimiter = ' ')
ibt = InitialBinaryTable.InitialBinaries(m1=dat_in.iloc[8].mass_1, m2=dat_in.iloc[8].mass_2, porb=dat_in.iloc[8].porb, ecc=dat_in.iloc[8].ecc, tphysf=13700.0, kstar1=1, kstar2=1, metallicity=0.0002)
bpp, bcm, initC, kick_info = Evolve.evolve(initialbinarytable=ibt, BSEDict={}, params='Params_RapidVink241220.ini')
```

[Params_RapidVink241220.txt](https://github.com/user-attachments/files/18334352/Params_RapidVink241220.txt)

@TomWagg  -- we should check this out becuase there is something in the Kiehl and Hurley prescription that catches this, but it _doesn't_ get caught in the Pfahl prescription

TomWagg:
This PR does two main things

1. Fix the debugging environment setup to handle the new file structure
2. Add conditions for massless remnants in the kick prescription to avoid NaNs

### NaN stuff

From #687 you can see how NaNs are created after a PISN, which caused hanging in the new Pfahl prescription. I have made two changes

- Add a condition to check for NaN sep values when seeing whether a system is bound (avoids the hanging)
- Add a condition for setting `vk` to a large number when `m1n=0` (since this could create infinite kicks in some prescriptions) - this avoids creating the NaNs in the first place

### Debugging stuff
`.vscode/tasks.json` and `src/cosmic/src/Makefile` needed some changes to paths to ensure everything is compiled in the right place. I also modified `debug/create_binary_in.py` and `src/cosmic/src/test_bse.f` such that all settings are accounted for. Now you can just run something like
```
python create_binary_in.py -f path/to/initC.h5 -b 42
```
Before triggering the debugging routine to run on a particular binary from an initC file (bin_num=42 in this case).

This will fix #687

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,7 +6,7 @@
             "type": "shell",
             "command": "make test",
             "options": {
-                "cwd": "${workspaceFolder}/cosmic/src"
+                "cwd": "${workspaceFolder}/src/cosmic/src"
             },
         }
     ]
--- a/debug/create_binary_in.py
+++ b/debug/create_binary_in.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import argparse
 
 BSE_settings = {'xi': 1.0, 'bhflag': 1, 'neta': 0.5, 'windflag': 3, 'wdflag': 1, 'alpha1': 1.0,
                 'pts1': 0.001, 'pts3': 0.02, 'pts2': 0.01, 'epsnov': 0.001, 'hewind': 0.5,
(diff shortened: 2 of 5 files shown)


katiebreivik (commented):
Thanks @TomWagg ! I think this all looks great. 

For the debugging, I'm wondering if we need to worry about connecting to the docs overhaul PR or if you think the two should play nicely if we merge into develop.

TomWagg:
I _think_ it should play nicely if we just merge!

Also let me know if you want to try merging the docs one in over Zoom and we can troubleshoot as we go if necessary :)

katiebreivik:
Ok! then I'll go ahead and merge this one in. I'll message separately about a docs hack!

PR #690, opened 2025-02-20
fix_WDrad

aliperego:
Fixed WD radius at line 2450 in evolv2.f by calling star() and hrdiag().


katiebreivik:
Thank you @aliperego !!

PR #695, opened 2025-03-27
Allow Independent sampler to return low/high-metallicity binary table for METISSE

d-maclean:
This PR updates the independent sampler to not truncate the input metallicity to the SSE limits $\big(1e-4 \le Z \le 3e-2 \big)$ if `metisse` is passed as the `stellar_engine` in `SSEDict`.

I also took the liberty of adding a user warning which informs the user when they request binaries from SSE which have an invalid metallicity. In such cases, the metallicity is truncated as before. (⌐■_■)


d-maclean:
It appears that the unit tests are failing because we're on cachev2, which was deprecated last month. For what it's worth, I'm able to build this commit error-free on my machine (x86-64), (python 3.9 & 3.10).

katiebreivik:
Hey @d-maclean! I updated the versions in the METISSE-integrate branch but you'll need to update your commit hash to use the new version of the workflows. See e.g. here: https://stackoverflow.com/questions/67868483/how-to-rerun-github-actions-workflow-when-the-pull-request-base-branch-updated

I think you should be able to run the following based on the stack overflow:
`git commit --no-edit --amend`
`git push -f`

katiebreivik (approved):
This looks great! Thanks @d-maclean

PR #696, opened 2025-04-06
Docs docker fix

katiebreivik:
Fixing tiny errors on the docs and docker builds:
- docs: restarting the binary failed because there was a merger that led to there not being the correct index access
- docker: our github actions workflows had out of date actions


katiebreivik:
gonna merge to see whether the fix works for docker; if it continues to fail, I'll debug in the docker workflow.

PR #699, opened 2025-04-12
Metisse integrate branch - fix default inifile and independent sampler from cosmic-pop

d-maclean:
Hello again,

I detected some issues while using `cosmic-pop` on the METISSE-integrate branch. Turns out I needed to modify the InitialBinaryTable calls to use the provided SSEDict values, or else the executable crashes due to changes made in my last pull request. :/

I also changed the default `acc_lim` and `alpha1` values in the `examples/Params.ini` inifile, to reflect a previous change which required these be 2-length lists for each component.


katiebreivik (approved):
👍

Issue #698, opened 2025-04-11
[Bug] Envelope mass is recorded wrong in the BPP during common envelope some cases


xevra:
https://github.com/COSMIC-PopSynth/COSMIC/blob/3fb7fadc04746e6eb848c7df24539e55c3c8b498/src/cosmic/src/comenv.f#L783

There are four sets of two cases in `comenv.f` where we check for supernovae and execute a `writetab` call in the event of a supernova during CE.

In what I'm calling "Case 3a", we have `switchedCE` is True, and another kind of switching going on.

The bpp entries for `mass0_X`, `menv_X`, `massc_X` written during CE are populated using three different methods:

1. `M01` / `M02` are passed into `comenv.f` directly using the appropriate `j` in `evolv2.f`
2. `masscX_bpp` is assigned in the internal logic of each "case" (E.g. Case 3a).
3. `menv_bpp` is passed as an array into `comenv.f` from `evolv2.f`, without any switching.

Katie and I verified that in most cases, the bpp records the correct behavior.

However, we verified that the envelope mass is recorded wrong in some cases. We believe this is Case 3a, for which `SwitchedCE` is true, and there is further switching is going on.

This could also affect Case 3b, but I am uncertain.

In addition to `menv_bpp`, we confirmed that `lumin` is also switched in these cases, and it likely effects other quantities passed to `comenv.f` as arrays.

The good news is that even when this is written wrong in the bpp, it is only written down wrong once, and is not propagated. Therefore, it appears  only as one bizarre line where things are switched near a CE event in the bpp.

I am planning to fix this in my branch, end eventually create a pull request. However, there's a lot of other stuff on my branch right now, so it might get fixed on develop faster if someone else looks into it.

xevra:
Notes from working on this issue:

Case 1a/1b only happen if the accretor is main sequence or stripped He.
Case 2a/2b happen if the accretor is a degenerate or giant.
Case 3a/3b appears to only happen when the accretor explodes.
Case 4a/4b are in the case of coalescence

M01, M1, MC1, AJ1, JSPIN1, KW1, formation1, bhspin1 are selected with star1
M02, M2, MC2, AJ2, JSPIN2, KW2, formation2, bhspin1 are selected with star2
star1 is the donor. star2 is the accretor

deltam_1 and deltam_2 are calculated before being passed into comenv
teff, radc, and ospin are calculated inside comenv (or a function)

Other quantities are passed as vectors.
[...]

PR #700, opened 2025-04-20
bugfix (issue 698)

xevra:
Bugfix addresses https://github.com/COSMIC-PopSynth/COSMIC/issues/698

Corrects a bug where sometimes quantities were reversed in bpp entries for supernova which occur during a common envelope.

-------------------------------------------------------------------------------

Quantities defined individually within the script (M01, M02, etc...) are labeled according to the `j1` and `j2` variables defined in `evolv2.f`. They are reversed in the `writetab` call for the BPP (following a supernova), placing them back in ZAMS order for consistency with `evol_type`

When `evol_type` is 15, the ZAMS primary is undergoing supernova. When 
[test_cosmic_CESN.py.gz](https://github.com/user-attachments/files/19826554/test_cosmic_CESN.py.gz)
it is 16, the ZAMS secondary is undergoing supernova. This requires reporting the ZAMS order of stellar objects in the bpp, rather than the order defined by `j1` and `j2`.

Quantities which are passed to `comenv.f` *as arrays* are always in the order of `ZAMS primary, ZAMS secondary`, and this is already the order for which they should be passed to writetab.

That order should not change as a result of switchedCE, as other quantities do.

I tested this using a script which identifies CE SN in the BPP files as one of the 8 mentioned cases in the issue, and verified that these changes correct the reporting.

I was able to test cases 1b, 2a, 2b, 3a, and 3b. If anybody finds a way to consistently report binaries in the bpp for cases 1a, or 4a/b, I would be happy to test them.

I am also happy to share my script for testing.

### DIFF
--- a/src/cosmic/src/comenv.f
+++ b/src/cosmic/src/comenv.f
@@ -21,6 +21,21 @@
 *     Update : P. D. Kiel (for ECSN, fallback and bugs)
 *     Date : cmc version mid 2010
 *
+*     Update : V. E. Delfavero (bugs in bpp reporting)
+*     Date : 20th, April 2025
+*
+* Note on indexing binary components:
+*
+* M01, M1, MC1, AJ1, JSPIN1, KW1, formation1, bhspin1 are selected with star1
+* M02, M2, MC2, AJ2, JSPIN2, KW2, formation2, bhspin1 are selected with star2
+*
+* deltam_1 and deltam_2 are calculated before being passed into comenv
+*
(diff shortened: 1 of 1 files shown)


katiebreivik:
Ok! I've verified that what a SN happens _during_ a CE, the switch is indeed incorrectly applied:
Screenshot 2025-05-06 at 11 52 32 AM
However, when installing the new branch, this problem is fixed:
Screenshot 2025-05-06 at 11 52 41 AM


Thanks @xevra!

katiebreivik (approved):
👍

PR #701, opened 2025-04-24
Added functions to save singles data

aliperego:
Handled match calculations for singles, modified test files for samples. Singles data are saved in the same file under keys "conv_singles", "bpp_singles", bcm_singles".


katiebreivik:
thank you @aliperego !

PR #703, opened 2025-05-20
Making Pre-supernova mass quantities accessible in the BPP file

xevra:
The goal of this pull request is to update the behavior of COSMIC so that the supernova entry in the BPP (evol_type 15 or 16) reflects the state of the star prior to `assign_remnant`.

The current behavior is:
mass0 is not updated during the `evol_type == 15 / 16` table entry.
menv is not updated during the `evol_type == 15/ 16` table entry.
massc is updated sometimes, but does not reflect the core mass of the star immediately prior to supernovae.

The desired behavior is:
mass0 is updated to contain the preSN total mass of the star in all subsequent bpp entries, including the `evol_type == 15 / 16` entry.
massc is updated to contain the core mass of the star prior to supernova in the `evol_type == 15 / 16` bpp entry.
menv is updated to contain the quantity `mcbagb - mc` from assign remnant during a Supernova explosion

Katie and I looked a `hrdiag.f` and determined that a Supernova of this kind can only occur when `massc` reflects a metallic core (not a Helium core). This enables us to print the metallic core mass of a star prior to Supernova, which is useful for updating supernova models, and comparing with other popsynth codes (specifically StarTrack).

The meaning of `mcbagb` is less clear, but can be interpreted as the Helium envelope, where the Hydrogen envelope can then be determined by calculating `mt - mc - mcbagb`. This is usually, but not always, close to zero.

We accomplish these objectives by introducing common block variables in `evolv2.f`, `assign_remnant.f`, and `comenv.f`, avoiding passing additional values to `hrdiag.f`, in order to avoid overhead in `hrdiag.f` which is called many times.


katiebreivik:
Will need to add in docs for the new columns but will merge this for now to clear the backlog!

katiebreivik (approved):
👍 !!

PR #704, opened 2025-06-01
Implement Disberg & Mandel 2025 model for natal kicks

TomWagg:
This PR implements the natal kick prescription from [Disberg & Mandel 2025](https://ui.adsabs.harvard.edu/abs/2025arXiv250522102D/abstract).

You can apply this prescription by choosing `kickflag = 5`.

Here's an example of the same population of 100,000 binaries evolved with the two different prescription. I plot the natal kick magnitude applied to each compact object in each case. You can still see the ECSN/USSN peak in the left side with a different distribution in the rest. You can see the probability of getting _very high_ kicks (>1100 km/s ish) is actually higher with this new one.

image


TomWagg:
I also haven't added any new tests for this. I was imagining running a population and checking the fraction of systems with kicks > 200 km/s or something and then seeing whether it properly changes for `kickflag = 1` vs. `kickflag = 5` but worry that may be slow to run if large, and prone to stochasticity if small. Open to ideas!

src/cosmic/src/kick.f
```diff
@@ -207,24 +213,33 @@ SUBROUTINE kick_pfahl(kw,m1,m1n,m2,ecc,sep,jorb,vk,sn,r2,
...
-             v(2*k) = s*SIN(theta)
- 25          continue
-          vk2 = v(1)*v(1) + v(2)*v(2) + v(3)*v(3)
-          vk = SQRT(vk2)
+                s = -2.d0*LOG(1.d0 - u1)
+                s = sigma*SQRT(s)
+                theta = twopi*u2
+                v(2*k-1) = s*COS(theta)
+                v(2*k) = s*SIN(theta)
+25          continue
```
michaelzevin, line 237: @TomWagg are we sure this continue statement is nested in the correct loop? It is inside the `do 25 k = 1,2` for the previous version.

michaelzevin (commented):
Looks good other than my question about the `continue`

src/cosmic/src/kick.f
TomWagg: I _think_ it doesn't matter? Since [the continue does nothing](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vn83/index.html) and the 25 label is the important part for the loop.

For most of the other DO...CONTINUE loops in the codebase the "continue" is un-nested (and I personally think it brackets where the loop is more clearly). I think the one in the original kick function was just a bit weird 🤷🏻

michaelzevin (approved):

TomWagg:
@katiebreivik I've added the tests for the Disberg kicks (and did Hobbs while I was at it). I just used the scipy `fit` functions instead of doing it the same way as `test_evolve` with creating your own histogram, but if there's a reason for doing it that way let me know and I can switch it over.

src/cosmic/src/kick.f
```diff
@@ -207,24 +213,33 @@ SUBROUTINE kick_pfahl(kw,m1,m1n,m2,ecc,sep,jorb,vk,sn,r2,
           xx = RAN3(idum1)
           xx = RAN3(idum1)
       else
+* If no pre-supplied kick magnitude, we draw a kick from a distribution
+* If the kickflag is 5 then use the log-normal distribution described
+* by Disberg & Mandel 2025
+          if(abskickflag.eq.5.and..not.ECSN_or_USSN)then
```
katiebreivik, line 219: It looks like `ECSN_or_USSN` is only used with Disberg+; just double checking: should we be applying a similar flag for the other kickflags?

katiebreivik (changes requested):
Hey @TomWagg  -- I just left one question re: `ECSN_or_USSN` in kick.f. Once we handle that, we can approve :-)

src/cosmic/src/kick.f
TomWagg: Yeah I purposely only applied this to kickflag 1 and 5 because the others in the settings don't seem to use ECSN. I figure we should leave it as is to avoid changing default behaviour?

katiebreivik: Got it -- and yeah, this makes sense!

katiebreivik (approved):
i think this is good to go! I'll move ahead with the merge. Love that the docs are updated so smoothly with the json!

PR #705, opened 2025-06-04
Singlesfix

katiebreivik:
- Remove the extra return from the initial sampler to handle the single stars properly. 
- Also cleaned up the warnings


katiebreivik:
Hey @aliperego  -- Can you check out the changes I made to see if this will still work with your `keep_singles`  use case?

PR #706, opened 2025-06-19
quick fix to prevent crash in point-mass sampler

carlrodriguez:
Very quick change to fix crash in point-mass clusters.  Changes single number in the CMC initial conditions.


katiebreivik:
👍

PR #708, opened 2025-06-29
Backport Disberg & Mandel 2025 Kick Prescription to METISSE-integrate

d-maclean:
This is a simple PR that cherry-picks @TomWagg's implementation of the Disberg & Mandel kick prescription into the METISSE-integrate branch. This pull also proposes bumping the minor version to 3.5.1 and notes the addition in the changes.

Examples of the new kicks in action for a population of neutron stars, featuring SSE & METISSE:
NS kicks SSE 06-29-25 positive

NS kicks METISSE 06-29-25 positive

I think it's worth noting somewhere that these new kicks are not usable with `kickflag < 0`, as the Kiel & Hurley 2009 subroutine doesn't include these changes. If there is demand to add this to the old kick options, I could see to that as well. :)

PR #709, opened 2025-06-30
Backport Disberg & Mandel 2025 Kick Prescription to METISSE-integrate

d-maclean:
This is a simple PR that cherry-picks @TomWagg's implementation of the Disberg & Mandel kick prescription into the METISSE-integrate branch. This pull also proposes bumping the minor version to 3.5.1 and notes the addition in the changes.

Examples of the new kicks in action for a population of neutron stars, featuring SSE & METISSE:
NS kicks SSE 06-29-25 positive

NS kicks METISSE 06-29-25 positive

I think it's worth noting somewhere that these new kicks are not usable with `kickflag < 0`, as the Kiel & Hurley 2009 subroutine doesn't include these changes. If there is demand to add this to the old kick options, I could see to that as well. :)

Addendum: I closed #708 because it targeted the `develop` branch by accident. My mistake. :grimacing:


TomWagg:
> I think it's worth noting somewhere that these new kicks are not usable with kickflag < 0, as the Kiel & Hurley 2009 subroutine doesn't include these changes. If there is demand to add this to the old kick options, I could see to that as well. :)

That's a good point, I hadn't thought about implementing it in the K&H version. Honestly I don't think we _should_ since I think people should be discouraged from that one (since it is wrong and only useful for reproducing old results). But it would be good to add a note in the docs that "-5" is not allowed.

d-maclean:
That's very reasonable. I went ahead and made note of that in Params.ini and the JSON docs.

katiebreivik (approved):
Thanks @d-maclean  and @TomWagg for your input. I think this is good to go!

PR #710, opened 2025-07-17
fixing the units for the central BH in CMC

carlrodriguez:
Quick fix to make sure the physical units in CMC are correct when including a central massive black hole.  Currently the BH is included in the unit conversions to physical units even though the internal code units do not include it.  This makes these consistent (both internal code units and physical units are only set by the cluster mass, not including the central BH).


katiebreivik:
merging this too @carlrodriguez !

PR #712, opened 2025-08-08
Preserve zpars when calling Evolve.evolve for METISSE; add Tauris+2015 ECSN prescription

d-maclean:
Hello,

This PR has two parts. First, I modified `zpars` to be intent(in,out) in evolv2, allowing us to capture it when we call `Evolve.evolve`. This is necessary for METISSE, as `zpars` is not internally preserved after the first call, and it is also not recalculated. Now, we simply feed it back in to subsequent `evolv2` calls, which should be fine as long as we don't allow changing metallicities mid-run.

Secondly, this adds the option to pass a negative number for `ecsn`. If this is done, `mc` is used to check for ECSN instead of mcbagb, allowing us to implement the [Tauris+2015](https://ui.adsabs.harvard.edu/abs/2015MNRAS.451.2123T/abstract) ECSN prescription. You simply pass the negative of your max C/O core mass for `ecsn`.


katiebreivik:
Hey @d-maclean  -- I want to do my own double check to see if all `ecsn` values are accounted for, but in the meantime, can you add an error check for if the SSEDict has metisse specified, that there can only be one metallicity per evolve call? This should be in the big error check function in utils.

d-maclean:
@katiebreivik Done! I had to place this in Evolve.evolve, as the error_check doesn't have access to the initial binary table.

d-maclean:
There is a related PR recently merged in [METISSE](https://github.com/TeamMETISSE/METISSE/pull/59#issue-3304645142) to force track reloading if `zpars` is all zeros. This is required if we call Evolve.evolve multiple times (like, in a jupyter notebook). I will push an update to the METISSE submodule to include that.

d-maclean:
This PR is being split up. Closing.

PR #713, opened 2025-08-18
Fixed tidal radius in CMC point mass sampler

mdodici:
In get_cmc_point_mass_sampler, changed default tidal radius from 1e13 to 1e6 virial radii. This change makes the function consistent with its documentation (and with get_cmc_sampler). This fixes a bug that arose when running isolated Plummer spheres in CMC, where routines to calculate the potential at the tidal radius would fail because the radius was too large.


katiebreivik:
Thanks for flagging this @mdodici! I'm looping in @carlrodriguez since I'm not quite as familiar with the CMC inner workings. 

@carlrodriguez  -- if you have merge capabilities feel free to merge if this looks good to you!

carlrodriguez:
This looks good to me (I fixed this in the realistic IC generator but not the point mass sampler; thanks for catching it!).

PR #714, opened 2025-08-21
COSMIC-METISSE interface

katiebreivik:
This PR removes the need to open any files during a fortran call, which makes multiprocessing possible and importantly, streamlines the eep load into a single thing you do at the beginning of the run. This is done through a new interface file: `c_m_interface.f90` which lives in `src/cosmic/`. There is also a co-PR over at TEAM-METISSE that completes the interface. 

There is a fair bit of runarounds that are created because Python and FORTRAN aren't always the closest friends, but this has been tested to reproduce binaries with the same outputs as the current develop branch over on METISSE and METISSE-integrate here on COSMIC.


katiebreivik:
Ok -- I think that at this point I need to wait for Poojan's updates that shunt the cosmic interface into it's own subroutine since the .mod isn't showing up in the compilation?

katiebreivik:
ok! Looks like the tests now finally pass. We really should also add some tests that work with METISSE data but i'm going to hold off on this for now.

PR #716, opened 2025-08-26
Capture & Recycle zpars for METISSE

d-maclean:
This PR adds a mechanism for capturing and recycling zpars from evolv2 when it is called with METISSE. This makes sure we keep zpars for remnant assignment, even if METISSE won't calculate it again between calls to Evolve.evolve.

### DIFF
--- a/src/cosmic/evolve.py
+++ b/src/cosmic/evolve.py
@@ -32,6 +32,7 @@ import warnings
 import os
 import sys
 import tqdm
+from functools import partial
 from pathlib import Path
 try:
     import multiprocessing
@@ -478,6 +479,9 @@ class Evolve(object):
             initial_conditions[i]["n_col_bcm"] = len(bcm_columns)
             initial_conditions[i]["col_inds_bcm"] = col_inds_bcm
 
+        # evolve one system to get zpars
+        _, _, _, _, _, zpars = _evolve_single_system(initial_conditions[0], None)
+
         # check if a pool was passed
         if pool is None:
             with MultiPool(processes=nproc) as pool:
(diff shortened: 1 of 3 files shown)


src/cosmic/evolve.py
```diff
@@ -636,7 +647,7 @@ def _evolve_single_system(f):
         _evolvebin.col.n_col_bcm = f["n_col_bcm"]
         _evolvebin.col.col_inds_bcm = f["col_inds_bcm"]
 
-        [bpp_index, bcm_index, kick_info] = _evolvebin.evolv2([f["kstar_1"], f["kstar_2"]],
+        [zpars, bpp_index, bcm_index, kick_info] = _evolvebin.evolv2([f["kstar_1"], f["kstar_2"]],
```
katiebreivik, line 650: Does this return change the order from line 685 below?

src/cosmic/evolve.py
```diff
@@ -737,30 +749,41 @@ def set_metisse_interface(path_to_tracks, path_to_he_tracks, IBT_Z):
...
+    # loop over the helium metallicity files to find the one that is the closest 
     # to the metallicity in the initial binary table
     met_dict_he_keep = None
     fmt_dict_he_keep = None
-    for m in met_files_he:
+    Z_idx_he = -1
+    for i, m in enumerate(met_files_he):
         met_dict, fmt_dict = utils.read_metallicity_and_format(m)
-        if met_dict['Z_files'] == IBT_Z:
+        if np.abs(met_dict['Z_files'] - IBT_Z) < Z_tolerance:
```
katiebreivik, line 772: this was part of a previous PR right?

src/cosmic/evolve.py
katiebreivik: I had some more time to check this and it's alright -- i was mixing up evolve calls.

src/cosmic/evolve.py
katiebreivik: forget this comment I think we are good -- though Poojan notes that we should use relative difference not absolute since absolute doesn't scale well with logarithmic metallicity grids.

katiebreivik (approved):
👍

Issue #692, opened 2025-02-23
Add  `Z_accuracy_limit` and `z_match_limit` to ini files


poojanagrawal:
In recent updates to COSMIC-METISSE,  METISSE varibales `Z_accuracy_limit` and `z_match_limit` were added 
to COSMIC. However, they are hard coded in get_COSMIC_input subroutine (defined in METISSE_utils.f90).

Can someone add them to the ini files and remove the hard coded parts?

TomWagg:
Fixed in #719

PR #717, opened 2025-08-27
Add COSMIC-METISSE interface to the independent sampler

d-maclean:
Currently, the new COSMIC-METISSE interface will cause a guaranteed crash when you call the independent sampler, as the METISSE interface is initialized in `Evolve.evolve`, not `sample`. This PR just adds a function call which sets the METISSE interface when you call the sampler.

### DIFF
--- a/src/cosmic/evolve.py
+++ b/src/cosmic/evolve.py
@@ -249,7 +249,7 @@ class Evolve(object):
         # NUMBER 1: PASS A DICTIONARY OF FLAGS
         BSEDict = kwargs.pop('BSEDict', {})
         SSEDict = kwargs.pop('SSEDict', {})
-        metisse_metallicity_tolerance = kwargs.pop('metisse_metallicity_tolerance', 1e-6)
+        z_accuracy_limit = kwargs.pop('METISSE_z_accuracy_limit', 1e-2)
 
 
         # NUMBER 2: PASS A PANDAS DATA FRAME WITH PARAMS DEFINED AS COLUMNS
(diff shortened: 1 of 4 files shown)


src/cosmic/evolve.py
```diff
@@ -760,13 +769,14 @@ def set_metisse_interface(path_to_tracks, path_to_he_tracks, IBT_Z, Z_tolerance)
     Z_idx_he = -1
     for i, m in enumerate(met_files_he):
         met_dict, fmt_dict = utils.read_metallicity_and_format(m)
-        if np.abs(met_dict['Z_files'] - IBT_Z) < Z_tolerance:
+        if abs(met_dict['Z_files'] - IBT_Z)/min(met_dict['Z_files'], IBT_Z) <= z_accuracy_limit:
```
katiebreivik, line 772: thank you for making it relative difference @d-maclean !!

src/cosmic/sample/sampler/independent.py
```diff
@@ -307,8 +307,10 @@ def get_independent_sampler(
 
     zsun = kwargs.pop("zsun", 0.02)
 
-    rad1 = initconditions.set_reff(mass1_binary, metallicity=met, zsun=zsun, SSEDict=SSEDict)
-    rad2 = initconditions.set_reff(mass2_binary, metallicity=met, zsun=zsun, SSEDict=SSEDict)
+    # TODO -- figure out how user can pass this
+    Ztol = 1e-6
```
katiebreivik, line 311: This hardcoded value should def be passed in by the user -- we should make it a part of SSEDict [Poojan's idea] with a default supplied. Maybe we set a default for the independent sampler and warn the user that it is being used.

katiebreivik (approved):
👍

src/cosmic/sample/sampler/independent.py
d-maclean: I'm will add the `z_accuracy_limit` to `SSEDict`. That way, the user can also pass it in the inifile. This would also address #692 . Default would be `1e-2`.

PR #719, opened 2025-09-01
Add `z_accuracy_limit` and `metisse_verbose` to COSMIC & inifile

d-maclean:
This PR is pretty straightforward. The `z_accuracy_limit` is both passed to `evolvebin.evolv2` and used in the `c_m_interface`. It should calculate the relative metallicity difference the same way METISSE does. Furthermore, I added `metisse_verbose`. although I'm not sure exactly what output that enables or where it is placed. This might be worth following up on. This should address #692.

### DIFF
--- a/src/cosmic/evolve.py
+++ b/src/cosmic/evolve.py
@@ -249,7 +249,6 @@ class Evolve(object):
         # NUMBER 1: PASS A DICTIONARY OF FLAGS
         BSEDict = kwargs.pop('BSEDict', {})
         SSEDict = kwargs.pop('SSEDict', {})
-        z_accuracy_limit = kwargs.pop('METISSE_z_accuracy_limit', 1e-2)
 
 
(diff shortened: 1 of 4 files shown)


src/cosmic/evolve.py
```diff
@@ -349,7 +349,7 @@ def evolve(self, initialbinarytable, pool=None, bpp_columns=None, bcm_columns=No
                     path_to_tracks=initialbinarytable['path_to_tracks'].iloc[0], 
                     path_to_he_tracks=initialbinarytable['path_to_he_tracks'].iloc[0],
                     IBT_Z=initialbinarytable['metallicity'].iloc[0],
-                    z_accuracy_limit=z_accuracy_limit
+                    z_accuracy_limit=1e-2
```
katiebreivik, line 352: Should this be `z_accuracy_limit` instead?

katiebreivik: Realized after chatting with @d-maclean that this is only specified if SSEDict is not passed, so it is ok!

katiebreivik (approved):
👍

PR #721, opened 2025-09-02
made fileio interface more general

poojanagrawal:
The python interface for reading files before could only handle specific kinds of files. Made it more general and also more robust. For example, there were issues with case between python and Fortran namelists. They are now dealt converting everything to lowercase in python. Also there were no defaults for the format and the metallicity namelists. Missing values would break the code.


katiebreivik (approved):
👍 Once the tests run I'll merge this in

PR #722, opened 2025-09-02
Update COSMIC-METISSE interface functions in utils / independent sampler

d-maclean:
This is a quick addendum to #721 which updates the function calls in the independent sampler to match the new signatures (i.e. one call for "path_to_tracks", and another optional one for "path_to_he_tracks").


d-maclean:
I've pushed another commit which reshapes an EEP's data in `read_MIST_track` to be 2-dimensional in the case that an eep has only 1 row. This shape should be invalid, but METISSE appears to catch and exclude these even if we don't.

PR #723, opened 2025-09-16
Smarter initC IO

TomWagg:
This PR will save ~1Kb per binary when saving an initC file without hugely affecting the runtime - that's a Gb per million binaries!

Identical columns are saved separately with a single row instead of N rows.
Hopefully the keyword args and docstrings are self-explanatory but typical usage is
```
...run COSMIC sim
save_initC("path/to/file.h5", initC)
loaded_initC = load_initC("path/to/file.h5")
```
instead of
```
.to_hdf() and pd.read_hdf()
```


katiebreivik (approved):
thanks @TomWagg !! Looking forward to smaller datasets!!

PR #739, opened 2025-12-17
Introducing `COSMICOutput`

TomWagg:
I went a little rogue and got carried away here but I think you'll like it @katiebreivik.

This PR introduces a new class in `cosmic.output.COSMICOutput`. This class lets you do things like:

- Instantiate based on evolve output `output = COSMICOutput(bpp, bcm, initC, kick_info, label="Latest run")
- Save and load from a file `output.save("file.h5"); new_output = COSMICOutput(file="file.h5")`
    - This will also track the COSMIC version and warns users about mismatches
- Index and mask the class to get specific binaries based on bin nums or masks
    - `randos = output[[0, 1, 2]]; randos = output[10:20]`
    - `massive = output[output.initC["mass_1"] > 10]
    - This will mask all of the relevant dataframes
- Re-run with new physics
    -  `alpha_5 = output.rerun_with_settings({'alpha1: 5}, inplace=False)`
- Plot all sorts of stuff
    - `output.plot_detailed_evolution(bin_num=0)`
        - This uses a bcm table, can pass kwargs to `plot_binary_evol` 
    - `output.plot_distribution(x_col='mass_1', y_col='mass_2', c_col='porb', when='initial')
        - This one uses either initC or bpp based on whether you say "initial" or "final", it can also do histograms if `y_col=None`

There's also a whole new tutorial in the docs that explains/demonstrates this all in more detail.

Happy holidays 🙃


katiebreivik (changes requested):
This is great! I think we need to add in a small change that handles [cosmic-pop](https://github.com/COSMIC-PopSynth/COSMIC/blob/develop/bin/cosmic-pop) output with the new class. Let's do it in a couple weeks.

TomWagg:
@katiebreivik this now has ``COSMICPopOutput`` which can read in COSMIC-pop outputs and convert them to individual ``COSMICOutput`` instances.

Ready to merge I think! :)

PR #742, opened 2026-01-16
Added martinez26 orbital period distribution

MarkGM02:
Introduced a new orbital period distribution `martinez26` which applies the `sana12` prescription to binaries with primary masses greater than or equal to 8 M⊙ (limiting the upper bound to 3000 days instead of 10^5.5), and the `raghavan10` prescription to binaries with less massive primaries.

Changed the name of the binary fraction based on Offner+23 to `offner23` which was named `offner22` originally.


docs/cosmic-settings.json
katiebreivik, line 1: Hey @TomWagg  -- This PR only modifies this file to update the docs to add a new model and fix a typo. I don't think this will require CMC review, but I'm not sure there's a way to avoid the CMC review flag for small changes like this. Do you have any ideas?

TomWagg: Hmm yeah that's tricky. The GitHub action basically is just detecting whether the file is changed but isn't doing anything clever about what's happening with the content. I'll have a think about making it more sophisticated

katiebreivik:
I think we have a bit of a plan for the CMC Flag moving forward based on offline discussions. I'm gonna go ahead and merge this one!

katiebreivik (approved):

PR #745, opened 2026-02-04
Added Seperated Helium and CO Core Masses to the Documentation

MarkGM02:
Minimal change so the documentation is up to date.
- Updated bpp/bcm column docs to replace `massc_*` with split core columns:
  - `massc_he_layer_*`
  - `massc_co_layer_*`
- Updated interactive inifile convergence options to use the `massc_co_layer_*` column names.


MarkGM02:
Also changed the table dimensions in ``docs/pages/output_info.rst`` so that the new long column names rendered without error!

MarkGM02:
Updated the definition of ``remnantflag`` so that its reliance on ``rembar_massloss`` is clearer!

katiebreivik (approved):
👍 Thanks @MarkGM02 

I'll merge this now!

PR #746, opened 2026-02-09
Update Params.ini

cferrus:
Added in the keep_singles line and a comment about needing to comment out the m2_min if running cosmic-pop in terminal due to an error requesting it. I thought it would be useful because I kept forgetting about it when I make a new Params.ini and I have to go through the 3 rounds of errors before it works.


TomWagg:
Thanks @cferrus! This actually highlights an issue - cosmic-pop and the sampler in general should allow the specification of m2_min=None as long as it's None. I'll push a fix for that too :)

PR #747, opened 2026-02-09
Update Params.ini

cferrus:
Changed the example for convergence limits. If used in the way that it was originally shown in the example, it wouldn't work.


TomWagg:
Thanks!

PR #750, opened 2026-02-12
Update independent.py

cferrus:
Changed the description of zsun


katiebreivik:
The metallicity kwarg is used for setting the radius in certain contexts so I'm going to close this one.

PR #751, opened 2026-02-12
Update independent.py

cferrus:
Changed met so it could intake "zsun" as a parameter because the wording of the description made it seem like that was an option. Set metallicity to zsun after zsun is defined.


katiebreivik:
The metallicity is supplied as Z=1-X-Y and fixed to a solar scale with zsun, so the user shouldn't be supplying 'zsun' as a metallicity. Instead, they should fix what the zsun assumption is (usually 0.02, 0.019, or 0.014), then supply the metallicity they want to run with.

PR #754, opened 2026-02-13
Update cosmic-pop

cferrus:
Removed the default for apply-convergence-limits because it is already defined in Params.ini. Otherwise, cosmic-pop throws an error about conflicting statements even if the user didn't specify anything on the command line about apply-convergence-limits. Also, fixed a minor typo.


katiebreivik:
Hey @cferrus  -- can you combine the PR's you've submitted into a single PR? Since they are mostly typo fixes it will be easier to keep track if they aren't spread across several different submissions.

katiebreivik:
This was fixed in #770 ; thanks for bringing it to our attention!

PR #755, opened 2026-02-16
Edge case bug fixes

### LINKED ISSUE #724: Massive massless remnants 😅
I've found a couple of massless remnants that have masses. I imagine this is a logging issue since the other evolution looks okay. I'll try using the debugging setup to investigate soon.

initC attached, bpp you can see here. It seems to be CE related?

[massive_massless_remnants.csv](https://github.com/user-attachments/files/22393972/massive_massless_remnants.csv)

Image

### LINKED ISSUE #725: Systems that unmerge
Similar to #724, this just seems like a logging issue, but some systems UNmerge ‼️
I had a sample of ~3000 of these ran with 3.4.17 and only 1 is still a problem with v3.6.1 so that's good 🙃 

Here's the initC and bpp

[unmergers.csv](https://github.com/user-attachments/files/22393991/unmergers.csv)

Image

### LINKED ISSUE #729: SNe during CE events don't log a disruption in BPP
When a system is disrupted, it should change its separation to -1, log disruption=1 in the kick_info table, _and_ add a BPP row that has evol_type == 11.

Attached are 84 examples where the latter of those three steps is not taken.

It seems to occur because the SNe happen during a CE. Additionally each system has two CEs so that is likely relevant.

[sn_during_ce__bpp.csv](https://github.com/user-attachments/files/23510465/sn_during_ce__bpp.csv)
[sn_during_ce__initC.csv](https://github.com/user-attachments/files/23510467/sn_during_ce__initC.csv)
[sn_during_ce__kick_info.csv](https://github.com/user-attachments/files/23510466/sn_during_ce__kick_info.csv)

TomWagg:
- Fixes #724 
- Fixes #725 
- Fixes #729 

I took a crack at some of the weirdos I've found recently with the GDB setup and I have fixed all of my test cases at least.

Changes are:
- Ensure disruptions are logged with ``evol_type==11`` when SNe occur during CEs
- Set ``tb=sep=0`` for cases where AIC caused a WD to explode and leave behind no remnant (``evolve_type==9``)
- Remove bug where ``kstar=15`` was assigned its previous epoch mass after merging during a CE and the merger product goes SN

To discuss:

1. For that last one I think I am correct to comment out these chunks of code, since Katie seems to have commented out an identical chunk in 2023 elsewhere in the code
2. For AIC of a WD -> should we be logging this in the bpp
3. Similarly, a CE -> merger -> SN is only logged as CE -> SN. I think we should have a row for the merger too no?


TomWagg:
@katiebreivik I recommend reviewing this before #756

katiebreivik (approved):
I did not, in fact, review this before the other PR, but I think we are good nonetheless. Thanks @TomWagg !

PR #756, opened 2026-02-16
Fixes to remnant/CO masses

### LINKED ISSUE #479: cmc has mass-dependent pts1 values should cosmic also have this?
We know that the setting for pts1 has a strong impact on the total runtime of a binary based on tests that @astroJeff has done. This is a big enough deal in cmc, which is way more computationally expensive, that it is important to dynamically set pts1 for massive/not massive stars. Maybe we should consider a similar type of dynamical setting for cosmic to optimize?

TomWagg:
This contains _several_ fixes to how CO core masses/remnant masses are handled. It also adds a new PISN prescription and windflag.

- Fixes:
    - Update ``mc_co`` and ``mc_he`` after adjusting ``mc = mcmax`` in ``hrdiag.f`` for stripped stars. This can be a fairly significant change, up to ~2 Msun. (Used to actually be up to 10 Msun because we added HeMS core mass growth)
    - Make ``evolv2.f`` actually log PISN (they get set to ``kw = 15`` so were missed previously)
    - Change ``loop = 40000`` (up from 20,000), otherwise 70 Msun stars run out of loops with new mass modifier
    - Changed all existing ``pisn`` prescriptions to use ``mc_tot`` instead of ``mcbagb`` for conditions
    - Update ``mc = mt`` for **ALL** remnants instead of just for certain flags
    - Only apply conversion from baryonic to gravitational mass once. Previous code **APPLIED IT TWICE!!** for core-collapse BHs! Create subroutine ``baryonic_to_ gravitational_mass``, which is called throughout
    - Ensure stellar type changes during hrdiag are logged to the bpp before SN occurs

- Additions/changes:
    - Linearly grow the core mass of HeMS stars from 0 to the value at the start of HeHG and check for core-collapse for all He stars (not just HeHG and HeGB)
    - Add ``dt_mass_modifier`` argument to ``Evolve.evolve()``.This lets you specify a list of tuples ``(m_low, m_high, mod)``, which multiplies ``pts1/pts2/pts3`` by ``mod`` for stars with primary star (only care about primary, conservative) ZAMS masses between ``m_low <= m_ZAMS < m_high``
    - Added a ``windflag = -1`` option to turn off stellar winds entirely
    - Added ``pisn = -4`` for Renzo+22/Hendriks+23 prescription for PPISN mass loss
        - With options ``ppi_co_shift`` and ``ppi_extral_ml``
    - Added ``fryer_mass_limit`` to limit remnant masses to the total core mass when using Fryer prescription

- Code cleanup:
    - ``assign_remnant`` no longer takes ``mc_tot`` as a parameter, just get it from the common block
    - WD chandrasekhar check use ``mc_co`` instead of ``mc``
    - Remnant flags 0-4 inclusive use ``mc_co`` instead of ``mc``
    - Got rid of ``mcx`` in ``assign_remnant`` in favour of clearer ``m_proto`` and ``m_FeNi`` to match the papers
    - [Very minor] Fryer Rapid was using <= instead of < everywhere

Here's a before and after with default settings.

image

And a comparison of PPISN differences:

(The spike in green in the right panel will go away with the LBV flag changes in an upcoming PR)

image


katiebreivik (approved):
This looks good to go based on our offline convo and the updated new settings page! Thanks @TomWagg!

PR #759, opened 2026-02-17
The previous comment was incorrect and was fixed to reflect what is in the code. 

julia-haynes:
The previous comment incorrectly states that the proto-core mass is set to 1. I changed the comment to reflect what is happening in the code, which is based on section 4.3 equation 18 in Fryer et al. 2012.

### DIFF
--- a/src/cosmic/src/assign_remnant.f
+++ b/src/cosmic/src/assign_remnant.f
@@ -165,7 +165,8 @@
 *
 * Use the "Delayed" SN Prescription (Fryer et al. 2012, APJ, 749,91)
 *
-*                    For this, we just set the proto-core mass to one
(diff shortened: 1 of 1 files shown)


TomWagg:
Nice catch @julia-haynes! Funny timing as well because I actually noticed that same thing in the PR I opened just a few hours ago and fixed the comment (it's buried in amongst all those changes I promise haha), so this should already get covered once we merge #756 :)

TomWagg:
p.s. If you're looking at `assign_remnant.f` in detail at the moment, you may want to wait until after Katie merges that PR (or checkout my branch) because I made quite a few changes/improvements to it

julia-haynes:
Sounds good! And I was looking at it in detail, thank you for letting me know!

PR #760, opened 2026-02-18
Added metallicity dependence to martinez26 and added martinez26_ecsn

MarkGM02:
Changed the primary mass from 8M⊙ to 6.8M⊙ which divides high and low mass stars for the martinez26 orbital period option. This cutoff mass was chosen since it is the minimum mass for an FeCCSN at COSMIC's lowest metallicity, when default winds are used.

### DIFF
--- a/src/cosmic/data/cosmic-settings.json
+++ b/src/cosmic/data/cosmic-settings.json
@@ -123,7 +123,11 @@
                     },
                     {
                         "name": "martinez26",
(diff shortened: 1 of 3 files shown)


katiebreivik:
@TomWagg mentioned that we might consider making this a metallicity-specific limit but I’m happy to merge with the inclusive 6.8 Msun limit for now once that one comment is addressed above.

TomWagg:
Yeah I think you could make a pretty simple prescription that accounts for the metallicity dependence. Here's a grid of single stars from COSMIC and a (very crude) (but not bad!) fit :)

image

TomWagg:
Or you could try fitting _any_ SN maybe?

Here the expressions are as follows:

For CCSN:
$M = 1.1 \log_{10}(Z) + 10.1$, but $M = 6.825$ for $Z < 10^{3}$

For any SN:
$M = 1.1 \log_{10}(Z) + 10.1$ at $Z > 10^3$, $M = 6.38$ for $Z < 3\times10^{4}$, and $M = 0.8 \log_{10}(Z) + 9.2$ in between. 


image

MarkGM02:
image

Implemented this using linear interpolation in log-metallicity between nearest grid points from my single star grids!

MarkGM02:
The mass cutoff is the FeCCSN minimum mass for `martinez26` and the ECSN mass for `martinez26_ecsn`.

katiebreivik:
thanks @MarkGM02 !

PR #762, opened 2026-02-18
Small fix to CO core masses for AGB stars

TomWagg:
In #756 we changed the conditions in ``assign_remnant.f`` to use ``mc_co`` instead of ``mc``. However, we didn't ensure that ``mc_co`` was set to ``mcmax`` if we overshot it. This meant that there was a timestepping issue that gave you a nonmontonic result of an SN occuring with smooth changes in initial masses and metallicity.

Here's the testing script:
[min_sn_crit.py](https://github.com/user-attachments/files/25401944/min_sn_crit.py)

Before:
image

After:
image


katiebreivik (approved):
Thanks @TomWagg !!

PR #765, opened 2026-02-19
Check for SNe immediately after CEs

### LINKED ISSUE #764: Occasionally SNe during CE do not log in bpp
I ran a simulation 10 million binaries and in 89 cases (it's pretty rare ;P) there was a SNe during a CE that wasn't logged in the bpp.

Here's the initC

[no_sn_log.csv](https://github.com/user-attachments/files/25422823/no_sn_log.csv)

And an example

Image

TomWagg:
Previous code didn't check for SNe after a call to `hrdiag` after a CE. This fixes #764.

Definitely check how I did this, I copied the block of code from the other hrdiag calls and made some slight adjustments to ensure that the evolve_type=8 is still logged each time. I _think_ it works (it fixes my weirdos definitely).

I've also switched the version 3.7.4 so you can release easily.

Oh and I also cleaned up the issue templates.


katiebreivik (approved):
this looks great -- thanks @TomWagg !

PR #767, opened 2026-02-25
Fixes for tides, mean_anomaly, and metallicity error checking

### LINKED ISSUE #766: COSMIC lets you run high metallicity without warning
I was able to run a COSMIC cluster above it's maximum supported metallicity without it returning a warning which led to unexpected behavior (i.e. some systems appearing at the beginning in the bcm file, but not in the end). See example of failing events:

Image

and attached initC file

[nans_initC.csv](https://github.com/user-attachments/files/25490832/nans_initC.csv)

It would be beneficial to have a warning so people don't run COSMIC in this unsupported way.

TomWagg:
A few small fixes here

- Prevent tides calculations based on NS/BHs
     - Missing this was previously leading to some BH + NS systems getting disrupted by tides 😅 
- Track mean_anomaly (save to kick_info) array even when a system is disrupted
- Throw an error if invalid metallicities are provided for evolution
     - This fixes #766 

I also made the debugging setup track the kicks as well, previously it just used random kicks.

@katiebreivik I bumped the version to v3.7.5. I don't think you've released 3.7.4 yet(?) but I think this is probably a different version now otherwise we'll hate ourselves in the future when 1000 things changed in a single version haha.


TomWagg:
CMC review bot doesn't need to worry here, the changes to ``evolve.py`` just check for metallicity. Removing label/comment.

katiebreivik (approved):
thanks twagg!

Issue #768, opened 2026-03-11
preSN masses are still being logged and written from comenv.f


katiebreivik:
While working through a bunch of merge conflicts, @xevra and I found that `comenv.f` is still using some prSN masses, e.g.:
https://github.com/COSMIC-PopSynth/COSMIC/blob/ffb858c3c70bc4225b1aed6d57ed3872bf4413bb/src/cosmic/src/comenv.f#L59
https://github.com/COSMIC-PopSynth/COSMIC/blob/ffb858c3c70bc4225b1aed6d57ed3872bf4413bb/src/cosmic/src/comenv.f#L319

@TomWagg  -- was this intentional? I think we may have totally spaced fixing this up with the other core mass changes?

xevra:
These were being used as globals (common) previously, but then the globals were removed in favor of an expanded bpp, and the preSN variables were removed in evolv2, assign_remnant, but without setting these in assign_remnant, they appear to be called uninitialized in comenv.

I am working on a fix, and should have it pushed to my branch today (and open a PR). However, the disrupt block at the bottom has me confused. I think that should be `menv_bpp(1)` and `menv_bpp(2)`, but am not 100% sure.

PR #770, opened 2026-03-11
Fixing comenv SN parameters, adding new wind/remnant/ppisn options

xevra:
This pull request:
1. Addresses issue https://github.com/COSMIC-PopSynth/COSMIC/issues/768
2. Adds the Fryer+2022 remnant mass prescription with fiducial fmix/Mcritnsbh parameters
3. Adds the "Weak" pulsational pair instability treatment from StarTrack (which still uses remnantflag to determine BH mass **after** reducing the mass due to a pulsation).
4. Adds Vink winds multiplied by 0.33
5. Adds the Bjorklund+2024 Wind mass loss power law
6. Adds the Krticka+2025 Wind mass loss power law.

Please check my treatment of the disruption event in comenv.f


xevra:
@katiebreivik @TomWagg

xevra:
Something is still wrong with the tests I use locally to test this version of COSMIC.

However, I'm having a hard time determining the issue by myself, and the comenv.f bug needs to be addressed, so I wanted to create the PR, and we can continue to test it before merging it.

TomWagg (dismissed):
This looks great @xevra! Not sure what @katiebreivik thinks here but I think it could be nice to make some of these things more flexible to the user. E.g. for Fryer+22 I think making f_mix a parameter that the user can specify would be great. And the 0.33 multiplier for the winds would be nice to expose to the user too.

I'll take a closer look at the PISN and the CE bit too after either it's an option to make this more flexible, or you both decide it's not necessary :)

I think bigger picture on the winds (not relevant to this specific PR), I think it would be very cool to make the whole thing modular and allow users to swap in different wind prescription for different phases (e.g. have WR winds and AGB winds and LBV winds with different options). At the same time @katiebreivik has pointed out that COSMIC-METISSE may make this obsolete anyway so maybe not an area we need to prioritise 🤷🏻

xevra:
@TomWagg I would love to add options for f_mix and the critical core mass for the Fryer+2022 prescription. However, I don't really know how to add inifile options that don't already exist. Do you have a list of steps that would be needed in order to accomplish this?

TomWagg:
Great! Yes I actually very recently created a docs page for this, so you can be the first to test it out! :)

https://cosmic-popsynth.github.io/COSMIC/pages/develop/new-settings.html

Let me know if anything is unclear, happy to help

xevra:
Finally got to this @TomWagg 

I tried to add the relevant options, but it's hung up on "positive_values".

Also, I updated the testing Params.ini file, but don't know how to update the hdf5 files. I'm familiar with h5py, but the structure is opaque to h5ls and similar methods.

```
Traceback (most recent call last):
  File "/Users/xevra/miniconda3/envs/basil/bin/cosmic-pop", line 221, in <module>
    utils.error_check(BSEDict, filters, convergence, sampling)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/xevra/miniconda3/envs/basil/lib/python3.13/site-packages/cosmic/utils.py", line 1202, in error_check
    raise ValueError(
        f"{flag['name']} must be one of {options} (you set it to '{user_val}')"
    )
ValueError: fryer_fmix must be one of ['positive_values', 1.0] (you set it to '2.0')
```

Have any suggestions?

Thanks,
Vera

xevra:
One more question: Do you have any idea how to *enforce* the Fryer+2022 critical core mass parameter?

I.e. If the core is above this value, explosion yields a black hole. If below, neutron star.

The remnant mass prediction is implemented here, but we're still using mxns to actually decide what kind of remnant we have, right?

So the question would be how to use the core mass rather than the total mass of the stellar object to make that determination when remnant=7 is set.

TomWagg:
Nice job @xevra!

- I think you want "positive values" not "positive_values" (right now it's checking for the string "positive_values" rather than the macro for "positive values" I think
- Good point on the HDF5 files, they are pandas dataframes saved so the structure is weird, I just added a function to the docs, pasting it below for convenience (it also changes Params.ini so you may need to make sure it doesn't duplicate the setting in there)
- On the determination of CO type, I would add to the conditions [here](https://github.com/COSMIC-PopSynth/COSMIC/blob/develop/src/cosmic/src/assign_remnant.f#L228-L236). Instead of checking the `mxns` parameter, instead first check if `remnantflag.eq.7` and then add your condition, _otherwise_ use `mxns`. Then update `data/cosmic-settings.json` to change the `mxns` description to highlight that it won't apply when `remnantflag` is 7 (probably also the `remnantflag` option 7 description too to make it extra clear). But @katiebreivik may have other opinions on this point?

```python
def add_setting_to_test_files(setting_name, setting_value):
    initC = pd.read_hdf("src/cosmic/tests/data/initial_conditions_for_testing.hdf5", key="initC")
    initC[setting_name] = setting_value
    initC.to_hdf("src/cosmic/tests/data/initial_conditions_for_testing.hdf5", key="initC", mode="a")

    initC = pd.read_hdf("src/cosmic/tests/data/kick_initial_conditions.h5", key="initC")
    initC[setting_name] = setting_value
    initC.to_hdf("src/cosmic/tests/data/kick_initial_conditions.h5", key="initC", mode="a")

    # this doesn't check if the setting is already in the file, ensure no duplicates
    with open("src/cosmic/tests/data/Params.ini", "a") as f:
        f.write(f"\n\n{setting_name} = {setting_value}\n\n")
```

xevra:
I seem to have fixed the default values macro! Thanks for the tips! Also updated the pandas table.

katiebreivik:
heads up @TomWagg: I'm going to dismiss the requested changes since I'm hoping to clean up PRs waiting in line for this one. I confirm that they've been incorporated!

katiebreivik:
Thanks for all the hard work on this @xevra  and for the review @TomWagg !

PR #771, opened 2026-03-26
Add multiprocessing capability to independent sampler

TomWagg:
I noticed that sampling was getting a bit slow once the number of binaries exceeded 10,000. I realised that COSMIC was doing sampling single-threaded. I don't think we need to do this?

So I've added options for `pool` and `nproc` (similar to `Evolve`) that can be used to parallelise the sampling. For a sample of 1 million binaries I can get a ~14x speedup when using 16 cores on my laptop (2.2s instead of 32s), I think this should approximately scale with the number of cores for large enough samples.

For now I set the minimum chunk size to 1000 binaries so that people don't overdo it on splitting up the sampling.

(Schwimmbad confuses me so feel free to swap that in for the basic multiprocesing pool I used @katiebreivik)


TomWagg:
You can test the speedups with this code :)

```
from cosmic.sample.initialbinarytable import InitialBinaryTable
from time import time
import numpy as np

final_kstar = [13, 14]

sampling_params = {
    'qmin': 0.0,
    'm2_min': 0.08,
    "binfrac_model": 1.0,
    "SF_start": 12000,
    "SF_duration": 0.0,
    "met": 0.02,
    "sampling_target": "size",
    "trim_extra_samples": False,
    "q_power_law": 0.0,
    "primary_model": "kroupa01",
    "ecc_model": "sana12",
    "porb_model": "sana12"
}

print("Sampling speed test time!!")
print("--------------------------\n")

for size in [10_000, 100_000, 1_000_000]:
    for nproc in [1, 4, 16, 32]:
        print(f"Sampling {size} binaries with nproc={nproc}...")

        start = time()
        ibt, mass_singles, mass_binaries, n_singles_req, n_bin_req = InitialBinaryTable.sampler(
            'independent', final_kstar, final_kstar, size=size, **sampling_params, nproc=nproc
        )
        end = time()
        print(f"    Took {end - start:.2f} seconds")
        print("    End results:")
        print(f"        Number of binaries sampled: {len(ibt)}")
        print(f"        Mass in singles: {mass_singles:.2f} Msun")
        print(f"        Mass in binaries: {mass_binaries:.2f} Msun")
        print(f"        Number of singles required: {n_singles_req}")
        print(f"        Number of binaries required: {n_bin_req}")
        print(f"        Primary mass percentiles: {np.percentile(ibt['mass_1'], [2.5, 50, 97.5])}")
    print()
```

katiebreivik (approved):
I think it's totally fine to just use the regular multiprocessing here. Thanks @TomWagg !

PR #777, opened 2026-04-29
Add `smt_periastron_check` setting

TomWagg:
Added `smt_periastron_check` setting, which allows the user to toggle whether to check for periastron collision during stable mass transfer. By default, we set this off, which changes the behaviour of previous versions since 3.6.0 (PR #694) which did check for contact at periastron during stable mass transfer. We set this to off as the radial expansion from SSE is likely overestimated, once we connect with COSMIC-METISSE this should likely be set to 1.


katiebreivik (approved):
👍

PR #782, opened 2026-05-27
updated error checking so Evolve.evolve will permit metallicities out…

d-maclean:
I changed a behavior in Evolve.evolve concerning metallicity bounds checking.

Previously, if you supplied binaries with a `z < 1e-4` or `z > 3e-2`, COSMIC would raise an exception due to the SSE boundaries. This change requires that METISSE not be the stellar engine for this exception to be thrown.


katiebreivik (approved):
🙏

PR #784, opened 2026-05-27
Error check on masses outside of the supplied grid for metisse use

katiebreivik:
Simple error check in evolve now to confirm that the masses evolved do not fall outside the limits of the tracks. 

flipped the order of the changelog


TomWagg:
Looks good!

image

PR #786, opened 2026-05-28
Add new column to `kick_info`, remove `bkick`, create what's new page dynamically

TomWagg:
This PR removes `bkick` and `kick_info_out`, so that CMC can just use `kick_info` directly now that we've added the `tphys` column.

Separately I added a method for dynamically generating a what's new page for the docs from the changelog.

I bumped the version to 4.1


TomWagg:
@claireshiye will care about this PR :)

TomWagg:
@katiebreivik 

image

katiebreivik (approved):
👍 

Thanks twagg!

PR #789, opened 2026-05-29
Skip METISSE mass bounds check for massless objects

d-maclean:
Currently, `Evolve.evolve` will throw an exception if you pass an initial binary table with `mass_1/mass_2 = 0` , because we have no zero-mass tracks. I added some logic to skip this if the mass provided is exactly zero.


katiebreivik (approved):
👍

PR #790, opened 2026-05-29
Added new kick prescriptions

astro-abhishek:
References for the kick prescriptions via `abskickflag = 7` and `abskickflag = 8`:

[1] Neutron Star Kicks by the Gravitational Tug-boat Mechanism in Asymmetric Supernova Explosions: Progenitor and Explosion Dependence, Janka H.T., 2017, ApJ 837(1):84, arXiv:1611.07562 [astro-ph.HE]. 
 -- Exact implementation matches Equation 2 from Chattaraj et al. 2026a.

[2] New constraints on the Bray conservation-of-momentum natal kick model from multiple distinct observations, Richards S. M., Eldridge J. J., Briel M. M., Stevance H. F., Willcox R., 2022, Published in: Mon.Not.Roy.Astron.Soc. 522 (2023) 3, 3972-3985.


astro-abhishek:
Here is a comparison of the existing natal kick prescriptions in COSMIC alongside the two new implementations introduced in this PR (`kickflag` 7 and 8).

Interestingly, Flags 1, 7, and 8 lead to similar looking distributions. If I understood correctly, Flag 1 (the standard COSMIC Maxwellian) gets its bimodal shape via hardcoded $\sigma$ values to differentiate between high velocity Fe core-collapse and low velocity ECSN/USSN. 

In contrast, the "physics-based" formulations in Flags 7 and 8 appear to produce this bimodality naturally. This was the conclusion of my [paper](https://ui.adsabs.harvard.edu/abs/2026ApJ...997...52C/abstract) as well. And the idea behind this PR based on what we discussed at CCA last month.

natal_kick_distributions

katiebreivik (approved):
👍 Thanks @astro-abhishek!!

Sorry for the delay on this!

PR #791, opened 2026-05-29
Clean up the API documentation so it properly links to the code

TomWagg:
This adds several docs pages with the API for all of the functions.

I also added the logo to the main and about pages.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,15 +27,6 @@ from cosmic import __version__ as cosmic_version
 sys.path.insert(0, os.path.abspath('.'))
 
 # -- General configuration ------------------------------------------------
-# This is the expected signature of the handler for this event, cf doc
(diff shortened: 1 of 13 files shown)


docs/conf.py
```diff
@@ -377,57 +371,38 @@ def setup(app):
...
-        return None
+        # use inspect to find the source code and starting line number
+        names = info["fullname"].split(".")
+        func = sys_mod
+        for name in names:
+            func = getattr(func, name)
 
-    modname = info['module']
-    fullname = info['fullname']
+        print(names, func)
```
katiebreivik: I think we don't want this print

katiebreivik (changes requested):

docs/conf.py
TomWagg: oop

katiebreivik:
👍

PR #792, opened 2026-05-30
Make sure docs cache rebuilds + organise tutorials + document `dt_mass_modifiers`

### LINKED ISSUE #758: Document ``dt_mass_modifiers``
I should write some docs about how you use this, perhaps in the same page as dynamic time resolution?

Just some quick examples of changing it and showing the effect on evolution/runtime/initC columns.

TomWagg:
I got carried away and this PR is actually 3 things sorry 😅 

# Docs caching

@katiebreivik pointed out that rebuilding the docs was super slow. I tracked down a few issues and no rebuilds are super fast locally, only edited files are changed :)

Why it was slow is basically two reasons. 1) The thing where BSEDict is generated dynamically meant that every time it dynamically inserted it then the file was marked as changed and removed from the cache. Now I first make it check whether the settings have actually changed and, if not, leave it in the cache. 2) In interface.rst we were including image directives that were referencing files made at fixed times and that also messed up the cache. Now [at]savefig handles all that.

(...let's not discuss who wrote the four files that were causing the issues 😇 )

# Tutorials organisation

I re-organised anything that was a tutorial into a clearer collection so it's easier to navigate. There's also a landing page at tutorials.rst that puts them all together neatly in a grid. Note this will change the URLs of these pages because they are in folders, but should only be an issue if someone has it hardcoded somewhere for some reason.

# `dt_mass_modifiers` docs

I wrote up a docs page for `dt_mass_modifiers` finally so this will fix #758


docs/conf.py
```diff
@@ -117,7 +120,7 @@
 
 # General information about the project.
 project = u'cosmic'
-copyright = u'2021, Katie Breivik'
```
TomWagg, line 120: @katiebreivik highlighting that I added my name to the bit that shows up at the bottom of each docs page (so that you can veto if you like, no worries either way! 🙂 )

katiebreivik: I think this is fair!

katiebreivik (approved):
This is fabulous! 👍

PR #801, opened 2026-06-15
Separate out constants into `consts.py`

TomWagg:
A pretty simple PR here that reduces the length of evolve.py by 100 lines - aren't for loops nifty? 🙃 

I've separated out the column constants into their own file. This will make setting them in `backpop` super simple. I've also updated the new settings docs (reduced number of things to change by 1 :D).

Oh and random small thing: CMC tests now use a temp directory for their sampling writeout rather than the directory you use to run tests. No more random input.hdf5 files :)


katiebreivik (approved):
This is great! Thanks @TomWagg

PR #802, opened 2026-06-16
Fix sampling when multiprocessing and using total_mass as your samplng target

TomWagg:
This fixes an issue added in v3.7.6 where sample fails if you have total_mass as your sampling target and use multiprocessing.


katiebreivik (commented):
👍 thanks for the quick turnaround!

PR #803, opened 2026-06-16
len(conv_filter_match) < 50 bug fix

astrogina:
Fixed bug where conv_filter_match was being overwritten in each step instead of being added to when it contained a nonzero number but less than 50 binaries.


katiebreivik (approved):
Thanks @astrogina! Love the extra log file writes; future users will def appreciate!

PR #805, opened 2026-06-17
Two-stage common envelope formalism from Hirai & Mandel 2022

TristanBruel:
Ciao!
Here is a proposed implementation in COSMIC of the two-stage common envelope formalism presented in [Hirai & Mandel 2022](https://iopscience.iop.org/article/10.3847/2041-8213/ac9519/pdf).
It basically relies on the idea that only the convective part of the envelope should be involved in the common envelope phase, and, if present, the radiative intershell more tightly bound to the core should only be lost through a stable mass transfer.
- When using SSE, it is based on fits from [Picker, Hirai & Mandel 2024](https://iopscience.iop.org/article/10.3847/1538-4357/ad4a5d/pdf) to compute the mass of the convective envelope and its binding energy.
- When using METISSE, the mass of the convective envelope is directly interpolated from the stellar tracks via mass_conv_envelope
It is worth noting that COSMIC technically already computes the mass of the convective envelope via the MRENV subroutine, but I could not find where these fits come from (I believe it is not from [Hurley et al 2000](https://academic.oup.com/mnras/article/315/3/543/972062?login=true)).
Cheers,
Tristan Bruel


katiebreivik:
Hey @TristanBruel  -- I am going to re-open this and review since I *finally* have an afternoon without teaching/meetings/course prep 🥳 

If you still want to close it post review, feel free!

src/cosmic/src/SSE/SSE_zfuncs.f
```diff
@@ -845,6 +845,94 @@ real*8 FUNCTION lmcgbf(mc,GB)
...
+      implicit none
+      integer kw
+      real*8 m,z
+      real*8 mcagbf,mifmcf
+      external mcagbf
+*
+* A function to evaluate the maximum mass of the convective envelope
+*       from Picker, Hirai & Mandel 2024
+*
+* I'm not sure how HG stars should be treated ...
```
katiebreivik: I think Picker+2024's Figure 1 shows that treating the HG is included in the fits so this should be ok.

src/cosmic/src/assign_remnant.f
```diff
@@ -697,8 +697,15 @@ SUBROUTINE assign_remnant_maltsev(mc, mc_tot, met, kidx, kw, mt)
          fallback = 1.0d0
          mt = mc_tot
       elseif (mc.gt.M2.and.mc.lt.M3.and.u_NS.lt.maltsev_pf_prob) then
-         fallback = maltsev_fallback
-         mt = (mc_tot - M_NS) * fallback + M_NS
+* Take linear relation from Ugolini+25
```
katiebreivik, line 700: Thanks for adding this in too!

src/cosmic/src/comenv.f
```diff
@@ -142,7 +184,44 @@ SUBROUTINE COMENV(M01,M1,MC1,AJ1,JSPIN1,KW1,
...
+              MENV = mconvenv(KW,M2,met,teff2,tmin,tset,AJ2,TM2)
+              mconvmax = menvmax(KW,M2,met)
+              LAMB2 = CELAMHE(M2,met,MENV,mconvmax)
+            ENDIF
+* if > 8 Msun, the envelope participating in the first stage is only the convective one
+* if < 2 Msun, the entire envelope participates in the first stage
+* linear interpolation in between
+            IF(M2.GE.8.0d0)THEN
+               m2endstage1 = M2-MENV
+            ELSEIF(M1.LT.2.0d0)THEN
```
katiebreivik: I think this should be M2 rather than M1

src/cosmic/src/comenv.f
katiebreivik: There is a sneaky extra CE all the way down here. I think it might suffice to put in a check on the flag to overwrite if the condition is met. What do you think?

src/cosmic/src/comenv.f
```diff
@@ -176,6 +255,63 @@ SUBROUTINE COMENV(M01,M1,MC1,AJ1,JSPIN1,KW1,
...
+      IF(CE2STAGEFLAG.EQ.1)THEN
+         SEPF = m1endstage1*m2endstage1/(2.D0*EORBF)
+* log end of the first stage with evolve_type=3
+         q1_bpp = m1endstage1/m2endstage1
+         q2_bpp = 1.d0/q1_bpp
+         rrl1_bpp = RC1/(RL(q1_bpp)*SEP_postCE)
+         rrl2_bpp = R2/(RL(q2_bpp)*SEP_postCE)
+         TB = (SEPF/AURSUN)*
+     &        SQRT(SEPF/(AURSUN*(m1endstage1+m2endstage1)))
+         CALL writetab(jp,tphys,3.d0,
```
katiebreivik: Can you add a: `if(using_cmc.eq.0)` check to this `writetab` call? Without it I think we might cause issues for CMC.

src/cosmic/src/comenv.f
```diff
@@ -176,6 +255,63 @@ SUBROUTINE COMENV(M01,M1,MC1,AJ1,JSPIN1,KW1,
...
+      ENDIF
+*
+* Two-stage energy formalism
+*
+      IF(CE2STAGEFLAG.EQ.1)THEN
+         SEPF = m1endstage1*m2endstage1/(2.D0*EORBF)
+* log end of the first stage with evolve_type=3
+         q1_bpp = m1endstage1/m2endstage1
+         q2_bpp = 1.d0/q1_bpp
+         rrl1_bpp = RC1/(RL(q1_bpp)*SEP_postCE)
```
katiebreivik: Is `SEP_postCE` defined here? Should this be `SEPF`?

TristanBruel:
Hi @katiebreivik, I wasn't sure if you were still interested in this and I think I might have messed up with my last commit today so I removed it.
But if you want to look into it I still think that it could be relevant for COSMIC.
Also I have added something in the Maltsev et al remnant mass prescription which has nothing to do with the two-stage common envelope formalism, feel free to ignore it.
If you need anything from my side let me know, I'd be happy to help

src/cosmic/src/comenv.f
katiebreivik: I think we also need a `switchedCE` check here similar to what is done at https://github.com/COSMIC-PopSynth/COSMIC/pull/805/changes#diff-baa2fb50d58bad1f0f99b5ea8b52f2bd8b653fc39508a17c618563f15966db20L294. If there is a secondary overflowing, then I think this will log incorrectly.

katiebreivik (changes requested):
Thanks for your patience on my massively delayed review @TristanBruel! I have had this on my radar to review since you opened it and have had a series of travel/workshops/teaching prep that have eaten every afternoon I planned to look through this PR carefully. 

I found a few small issues that shouldn't be too tough to fix. The main thing to fix is handling of our `switchedCE` case where the donor can be the secondary rather than the primary and hiding the `writetab` call inside a `using_cmc` check so we don't blow up the output of CMC. I tried to link to an example of how a `switchedCE` write is handled later down in the file, let me know if you have questions for how to handle this!

Lastly -- thanks for the new Malstev flag behaviour. It will be nice to have that extra option. 

Getting this merged will remain at the top of my to do's so I'll keep an eye to help bring this to the finish line!

src/cosmic/src/comenv.f
katiebreivik: This has been percolating in the back of my mind... I'm not 100% sure if the new code up above handles the `SEPF` for this case. It might be necessary to add the twostage results down here as well, but I hope that most of the necessary information is already calculated.

src/cosmic/src/SSE/SSE_zfuncs.f
TristanBruel: Actually I was not super confident about this part so I checked how it is implemented in COMPAS. They have different options to treat the nature of the envelope for HG stars (either fully radiative, or the user can choose a temperature threshold to transition from radiative to convective, or a threshold on the fraction of the convective mass with respect to that of the envelope). So I decided to change the code and implement the fitting formulae from Hurley et al. 2000. I don't know at this point if this makes a big difference, I could try to run some tests later on.

src/cosmic/src/comenv.f
TristanBruel: Actually this whole writetab call is not necessary, I added it to help keep track of the physical values between the two stages of this new formalism, and to make sure that it was indeed running correctly. I leave it for now with the addition of the condition if(using_cmc.eq.0) and the corrections on SEPF, but it can be entirely removed in the future.

src/cosmic/src/comenv.f
TristanBruel: Actually there is no need to re-assign these values here in this version of the code, it is totally a mistake from my part. 
I had already moved everything up to make it easier to define the final orbital separation depending on the flag CE2STAGEFLAG. I clarified that part (line 251 to 357) and removed that now unnecessary part in the case of degenerate or giant secondary.

TristanBruel:
Hi @katiebreivik, I checked your comments during the weekend and implemented the proposed corrections. It was all quite straightforward so I hope I didn't mess it up.
The only point with some real astrophysical impact is the treatment of Hertzprung Gap stars, and in particular the structure of their envelope. I wasn't really satisfied with the previous version, which basically treated them as Giants, so instead I implemented the fitting formulae from Hurley et al. 2000. I'm not familiar enough with detailed stellar evolution to know if it's more motivated, but after running a couple of tests it for sure makes a significant difference for the final state of binaries where a HG star triggers a common envelope phase.

src/cosmic/src/SSE/SSE_zfuncs.f
katiebreivik: Ok cool! Thanks for doing the extra work there! I think I'll make an issue so we don't forget to check that behavior

src/cosmic/src/comenv.f
katiebreivik: 👍

src/cosmic/src/comenv.f
katiebreivik: I think it might be nice to keep the extra so that the user sees the impact of the prescription, so I like keeping it in for now.

src/cosmic/src/comenv.f
```diff
@@ -120,17 +127,56 @@ SUBROUTINE COMENV(M01,M1,MC1,AJ1,JSPIN1,KW1,
...
+      IF(CE2STAGEFLAG.EQ.1)THEN
+         IF (using_SSE.eq.1) THEN
+* Estimate the mass of the convective envelope 
+* from the fits in Picker, Hirai, & Mandel 2024
+           met = 10**(LOG10(zpars(14))/0.4)
+           reagb = ragbf(M1,LUMS(7),zpars(2))
+           tmin = 1000.d0*((1130.d0*LUMS(7)/
+     &                       (reagb**2.d0))**(1.d0/4.d0))
+           tset = tonset(tmin,met)
+           teff1 = 1000.d0*((1130.d0*lumin(1)/
```
katiebreivik: I think the luminosity and radius should be L1 and R1 here to handle whether there are `switchedCE`

PR #807, opened 2026-06-22
Ensure random seeds are unique for multiprocessed sampling

TomWagg:
The title tells you what's up mostly, I messed up the sampling in v3.7.6 because multiprocessing workers need to be initialised with their own random seeds. I offset each worker by the process ID now so all should be fine. I've also added a test to ensure this doesn't happen again :)


katiebreivik (approved):
👍

PR #808, opened 2026-06-24
5 new config. figures (pts, neta, hewind, zsun, and windflag)

lukewilner:
.py files for pts, neta, hewind, zsun, and windflag


katiebreivik (changes requested):
Thanks for adding these @lukewilner! I have a few quick requests:

1) For zsun, we aren't sure that we need these figures since zsun shouldn't have a large range. Can you remove this file?

2) For windflag, can you add an additional panel that shows the same thing as the leftmost panel, but with the x/y axis on a log10 scale as an inset? An example is Figure 8 in this paper: https://iopscience.iop.org/article/10.3847/1538-4365/ac416c/pdf

3) For each figure, can you remove the title at the top in the small text? 

Thank you!

katiebreivik (approved):
Thanks @lukewilner !

PR #820, opened 2026-08-03
low-mass binary fraction in multidim sampler

cheyanneshariat:
## Summary
I did a small change to multidim.py to:
- set the absolute binary fraction at 0.08 solar masses to 20%, interpolating in log primary mass to the existing multidim prediction at 0.8 solar masses

Made a  regression test and update the seeded sampler expectations

This removes the previous assumption of zero binary fraction at the hydrogen-burning limit. The 20% endpoint is motivated by the observed nonzero multiplicity of very-low-mass stars (e.g. Winters et al. 2019, arXiv:1901.06364).

## Tests

- `pytest -q src/cosmic/tests/test_sample.py`: 28 passed
- `pytest -q src/cosmic/tests`: 113 passed
- `git diff --check`

### DIFF
--- a/src/cosmic/sample/sampler/multidim.py
+++ b/src/cosmic/sample/sampler/multidim.py
@@ -32,6 +32,31 @@ __credits__ = "Scott Coughlin <scott.coughlin@ligo.org>"
 __all__ = ["get_multidim_sampler", "MultiDim"]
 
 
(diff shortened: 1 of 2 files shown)


katiebreivik:
Thank you @cheyanneshariat !!

PR #821, opened 2026-08-04
Docs button

katiebreivik:
tiny pr that fixes the dead docs button pointers introduced by @TomWagg's nice reorg of our tutorials.


katiebreivik:
whoa! not sure I understand the 17 other commits but everything else seems to check out.

PR #823, opened 2026-08-07
Fixed a second typo in independent.py's criterion for Roche lobe overflow

nweatherford:
It turns out there were two typos that largely canceled each other out for most cases, but not always. My first commit only fixed one of the typos. This second fix is more important because implementing only one of the two fixes is worse than implementing neither! See further description below. I've also successfully tested the version with both fixes in CMC.

Also note I also slightly adjusted the notation in the code and I use the new notation when describing my changes below. Specifically, I renamed the former parameters q and RL_fac to be q1 and RL_fac1, respectively. So whenever I refer to "q" or "RL_fac" below, I am NOT referring to the old parameters. Rather, this means my statement applies to both q1 and q2, or both RL_fac1 and RL_fac2, respectively.

#########################################
The first typo (fixed in my first commit) was that when computing RL_fac from Eggleton (1983), the log term  was missing parentheses around q's 1/3 exponent. So the code was calculating that term as log(1+q/3) when it was supposed to be log(1+q**(1/3)). The typo meant that for q < 5.2, RL_fac would be overestimated: by ~46% for q=1, by 215% for q=0.1, and by arbitrarily much as q–>0. For q > 5.2, RL_fac would instead be underestimated, but only by as much as ~12% (for q~40), converging to the correct value as q–>infinity.

#########################################
However, there was also an additional typo I missed in my first commit, which I fix here instead. This second fix is more urgent because implementing only one of the two fixes is worse than implementing neither! In the Eggleton (1983) formulas, q = m_donor/m_accretor and RL_fac is the donor's Roche lobe (RL) radius over the binary's pericenter a(1-e).

In the old code, the mass ratios were defined to be (q1, q2) = (mass2/mass1, mass1/mass2). And the criteria for RL overflow was set to be twice the maximum of (rad1 / RL_fac1) and (rad2 / RL_fac2), i.e. the individual criteria for when the donor is star 1 or star 2, respectively. Note this is inconsistent with the definition of (q1, q2). I therefore redefined the mass ratios to be (q1, q2) = (mass1/mass2, mass2/mass1).

#########################################
I suspect the reason neither typo was noticed before (despite each one introducing a significant change to a_min) is that they incredibly managed to mostly cancel each other out. Under the constraint that (rad1, rad2) = (mass1**0.8, mass2**0.8)---i.e., the mass-radius relation for low-mass stars---the old code (before fixing either typo) would have underestimated the correct a_min (fixing both typos) by at most ~31%, occurring at q=1. (A maximum ~3% overestimate would have occurred near q=2400 or 1/2400, converging to 0% in the limits q—>0 or q—>infinity).


katiebreivik (approved):
👍

PR #826, opened 2026-09-10
Extend python/numpy/scipy version support and CI fixes

### LINKED ISSUE #825: Update to support numpy >=2.4 and scipy >= 1.17
Please update the package dependencies to use more recent versions of python, numpy, scipy, and pandas. Thanks!

mcdigman:
Closes #825

## Summary
- Extend support for python to 3.14, numpy to 2.5, scipy to 1.18
- Make several undeclared dependencies explicit (`dill`, `beautifulsoup4`/`bs4`)
- Added an explicit `extras` optional dependency group for the optional `legwork`
- Docs update re: the `extras` optional dependency
- Remove obsolete python 2.x compatibility code (including `six`, `importlib_resources` dependency)
- Replace obsolete/deprecated AST node references
- Upgrade dependency floors to versions that exist on pypi (`h5py` and `tqdm`)
- Add `lowest` CI job
- Added smoke test to CI to ensure built wheels work
- Add Python 3.14 to test matrix
- Add `data/__init__.py` to meson so editable installs work
- Removed unused dependency `importlib-metadata < 5.0` from the `test` extra
- Bumped `schwimmbad` floor to `>=0.4.0`
- `cibuildwheel` config migrated from workflow env vars into `[tool.cibuildwheel]` in `pyproject.toml`
- `CIBW_SKIP` gained `*t-*`, making the exclusion of free-threaded wheels explicit — `CIBW_BUILD`'s `cp3XX-*` patterns already excluded them, since those globs don't match the `cp3XXt-*` identifiers; let me know if that is not intended.
- fix the broken coverage CI job
- add unit tests for every uncovered line touched
- Fixed a circular import chain and removed an unused circular import

Linux CPython 3.14 wheels need to use a different build environment (`manylinux_2_28`), but the old build environment (`manylinux2014`) can still be used for 3.13 and prior, because upgrading it would break backwards compatibility on legacy systems.

## Version floor compatibility issue
The `lowest` CI job would have used `schwimmbad==0.3.1`, which implicitly requires `setuptools<82` because `pkg_resources` is removed after that. However, there is currently no declared setuptools requirement in pyproject.toml, and adding a new explicit dependency requirement `setuptools<82` in pyproject.toml would likely break/unnecessarily constrain some existing more modern working builds. 

Instead, I bumped the version floor to `schwimmbad>=0.4.0` to remove the setuptools compatibility issue. 

Alternatively, the  `lowest` job could just pin `setuptools<82` without stating that as a requirement in pyproject.toml, which would work but create a minor risk that some users could accidentally have a non-functional install with a set of versions that was not prohibited by pyproject.toml.

Let me know if you prefer an alternate approach.

Note: this PR did not *cause* the issue, the addition of the `lowest` CI job *detected* an existing issue with the declared dependency contract, which is its purpose.

## Coverage CI job fix
I noticed that the coverage CI is comparing every new PR to a 3 year old commit before the switch to `meson`, so it *always* fails. I added a fix (upload the coverage report on push to `develop`), which should hit on future PRs or reruns *after this PR is merged*; until then it will still be broken. The coverage job was also always reporting 0% for every single non-test python file due to the `src` tree, which should also be fixed now. 

The 0% came from installing non-editable; --cov=./src/cosmic measured a source tree that never executed. That requires --no-build-isolation (and preinstalling the build deps), because meson-python bakes the ninja path from pip's ephemeral build environment into the editable loader, and that directory is deleted after install — the first import cosmic would otherwise fail with FileNotFoundError.

## Packaging Note
Note that, currently, meson builds all of the tests *into the wheel at package root*, e.g. `cosmic/test_compatibility.py`, rather than `cosmic/tests/test_compatibility.py` or simply not building the tests into the installable wheel. I didn't change that because don't know if that is on purpose for some reason, but if you don't want to build the tests into the wheel, say so and I can make meson not build the tests into the wheel or at least build them into `cosmic/tests`.

## AI Use Disclosure
The fixes were a mix of manual work (inspecting dependency floors, finding unused or undeclared dependencies, deciding to add new CI jobs) and fixes guided by Codex GPT-5.6 Sol. The changes were cross-reviewed by Claude Opus 5.0 and a separate Codex GPT-5.6 Sol agent. I reviewed all the changes.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -3,7 +3,11 @@
 
 name: Unit Test COSMIC
 
-on: [pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - develop
 
 jobs:
   build:
@@ -47,20 +51,24 @@ jobs:
       run: |
         # upgrade apt-get and install system dependencies
         sudo apt-get update
-        sudo apt-get install gfortran swig libhdf5-serial-dev meson python3-dev
+        sudo apt-get install -y gfortran swig libhdf5-serial-dev python3-dev
 
         # upgrade pip and install build/lint tooling
-        python -m pip install -U pip setuptools wheel packaging flake8
+        python -m pip install -U pip setuptools wheel packaging flake8 \
+          "meson-python>=0.15" "meson>=1.3" ninja \
+          "numpy>=2.0,<2.3; python_version<'3.14'" \
+          "numpy>=2.3.2,<2.6; python_version>='3.14'"
 
     - name: Install package with highest-compatible dependencies
       if: matrix.dependency-set == 'highest'
       run: |
-        python -m pip install .[test]
+        python -m pip install --no-build-isolation -e .[test]
 
     - name: Install package with lowest direct dependencies
       if: matrix.dependency-set == 'lowest'
       run: |
-        python -m pip install -c ci/constraints-lowest.txt .[test]
+        python -m pip install -c ci/constraints-lowest.txt "numpy>=2.0,<2.3"
+        python -m pip install --no-build-isolation -c ci/constraints-lowest.txt -e .[test]
         
     - name: Test with pytest
       run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -7,12 +7,16 @@ on: [pull_request]
 
 jobs:
   build:
-  
+    name: Python ${{ matrix.python-version }} (${{ matrix.dependency-set }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
+        dependency-set: [highest]
+        include:
+          - python-version: '3.10'
+            dependency-set: lowest
 
     steps:
     - name: Checkout repository with submodules
@@ -34,11 +38,12 @@ jobs:
         # This path is specific to Ubuntu
         path: ~/.cache/pip
         # Look to see if there is a cache hit for the corresponding requirements
-        key: ${{ runner.os }}-pip-${{ hashFiles('pyproject.toml') }}
+        key: ${{ runner.os }}-py${{ matrix.python-version }}-pip-${{ matrix.dependency-set }}-${{ hashFiles('pyproject.toml', 'ci/constraints-lowest.txt') }}
         restore-keys: |
-          ${{ runner.os }}-pip-
+          ${{ runner.os }}-py${{ matrix.python-version }}-pip-${{ matrix.dependency-set }}-
+          ${{ runner.os }}-py${{ matrix.python-version }}-pip-
           ${{ runner.os }}-
-    - name: Install dependencies and package
+    - name: Install system and build dependencies
       run: |
         # upgrade apt-get and install system dependencies
         sudo apt-get update
@@ -47,8 +52,15 @@ jobs:
         # upgrade pip and install build/lint tooling
         python -m pip install -U pip setuptools wheel packaging flake8
 
-        # install cosmic together with its test dependencies
+    - name: Install package with highest-compatible dependencies
+      if: matrix.dependency-set == 'highest'
+      run: |
         python -m pip install .[test]
+
+    - name: Install package with lowest direct dependencies
+      if: matrix.dependency-set == 'lowest'
+      run: |
+        python -m pip install -c ci/constraints-lowest.txt .[test]
         
     - name: Test with pytest
       run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
 
     steps:
     - name: Checkout repository with submodules
@@ -25,7 +25,7 @@ jobs:
         git submodule set-branch --branch develop src/cosmic/src/METISSE
         git submodule update --remote src/cosmic/src/METISSE
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
     - name: Cache pip
--- a/docs/pages/tutorials/adaptive/basics.rst
+++ b/docs/pages/tutorials/adaptive/basics.rst
@@ -198,7 +198,7 @@ These functions focus on double compact objects (DCOs) and are suitable for grav
 
 .. note::
 
-    The ``merging_dco`` function requires the `LEGWORK python package <https://legwork.readthedocs.io/en/latest/>`_ to compute the merger time.  If you do not have LEGWORK installed, ``merging_dco`` will raise an error.
+    The ``merging_dco`` function requires the `LEGWORK python package <https://legwork.readthedocs.io/en/latest/>`_ to compute the merger time.  If you do not have LEGWORK installed, ``merging_dco`` will raise an error. It can be installed using ``pip install cosmic-popsynth[merging-dco]``
 
 .. code-block:: python
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,15 +22,15 @@ classifiers = [
     "Programming Language :: Python",
 ]
 dependencies = [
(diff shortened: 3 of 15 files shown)


.github/workflows/python-package.yml
```diff
@@ -3,16 +3,24 @@
...
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
+        dependency-set: [highest]
+        include:
+          - python-version: '3.10'
+            dependency-set: lowest
```
TomWagg, line 23: I haven't seen this before - could you explain more about the lowest dependency set part? Am I reading this correctly that it not only using 3.10 but also the minimum matching versions of the packages in that version?

docs/pages/tutorials/adaptive/basics.rst
```diff
@@ -198,7 +198,7 @@ These functions focus on double compact objects (DCOs) and are suitable for grav
 
 .. note::
 
-    The ``merging_dco`` function requires the `LEGWORK python package <https://legwork.readthedocs.io/en/latest/>`_ to compute the merger time.  If you do not have LEGWORK installed, ``merging_dco`` will raise an error.
+    The ``merging_dco`` function requires the `LEGWORK python package <https://legwork.readthedocs.io/en/latest/>`_ to compute the merger time.  If you do not have LEGWORK installed, ``merging_dco`` will raise an error. It can be installed using ``pip install cosmic-popsynth[merging-dco]``
```
TomWagg: I don't think it's worth having a separate extra for this, stating ``pip install legwork`` seems sufficient to me

src/cosmic/sample/stroopwafel/presets.py
```diff
@@ -31,7 +31,7 @@ def merging_dco(kstar_1, kstar_2, max_merge_time=13.7):
         from legwork import evol
     except ImportError:
         raise ImportError("The 'merging_dco' preset requires the LEGWORK package. "
-                          "Please install it with 'pip install legwork' to use this preset.")
+                          "Please install it with 'pip install cosmic-popsynth[merging-dco]' to use this preset.")
```
TomWagg: See above

src/cosmic/utils.py
```diff
@@ -35,11 +35,7 @@
 import h5py as h5
 import re
 
-import sys
-if sys.version_info >= (3, 9):
-    from importlib.resources import files as io_files
-else:
-    from importlib_resources import files as io_files
+from importlib.resources import files as io_files
```
TomWagg, line 38: Not sure whether @katiebreivik wants to continue support for below v3.9 of Python, given it's not in our tests I think it's reasonable to drop it though.

pyproject.toml
```diff
@@ -22,16 +22,17 @@ classifiers = [
     "Programming Language :: Python",
 ]
 dependencies = [
-    "numpy >= 2.0, < 2.3",          # if you change numpy versions, also change it in (1) the build-system section below and (2) the .github/workflows/build_wheels_and_publish.yml file
```
TomWagg, line 25: I think this comment is still necessary, at least it looks like the python-package.yml still has a lot of the specific numpy version hanging around so this warning is still useful. It's likely worth adding to this comment that these values change for more recent python versions.

pyproject.toml
```diff
@@ -47,20 +48,41 @@ doc = [
...
 ]
 test = [
     "pytest",
     "pytest-cov",
     "coverage",
-    "importlib-metadata < 5.0",
     "wheel",
 ]
-all = ["cosmic-popsynth[doc, test]"]
+merging-dco = ["legwork"]
```
TomWagg: See above, I don't think we need this.

TomWagg (changes requested):
Hi @mcdigman - thanks for the PR! @katiebreivik will probably want to look at this too but I thought I'd chime in since I was looking at the versions recently.

I think this looks great. Overall comments are just that I don't think LEGWORK needs to be defined in its own "merging-dco" section, it's just a single package so I think just referencing that is enough.

I'm particularly excited about the coverage fix, that's been bugging me for a while.

@katiebreivik we should also probably drop 3.10 support once we hit October (see https://devguide.python.org/versions/) and that should simplify a lot of this.

.github/workflows/python-package.yml
mcdigman: Exactly; it is a unit test that specifically tests whether the lowest possible combination of dependencies that the pyproject.toml _claims_ should work actually does work. In my experience, it is the most sensitive individual test in the matrix for compatibility breaking changes in the PR (like accidentally using a method that wasn't introduced until a later version, in which case you would either not use that method or bump the claimed floor). 

Highest overall is more sensitive to upstream compatibility-breaking changes; I feel like it is relatively unusual (though not impossible) for one of the middle unit tests in the matrix to be the only one that breaks in any way from a PR.

katiebreivik: 👍

docs/pages/tutorials/adaptive/basics.rst
mcdigman: I don't disagree necessarily; it could also be called something else less specific like `misc` maybe, in case you add other optional dependencies; some projects have a policy of documenting every dependency the project _might_ need at least somewhere in the pyproject.toml

katiebreivik: I like the idea of creating a `misc` for optional dependencies since we are also adding other post-processing modules like [eventlab](https://github.com/COSMIC-PopSynth/eventlab). 

@mcdigman  -- do you mind updating to `misc` instead of `merging-dco` here?

src/cosmic/sample/stroopwafel/presets.py
katiebreivik: update with `extras` here too

src/cosmic/utils.py
mcdigman: Note several existing pyproject.toml dependency floors already don't resolve on <3.10 ([SciPy 1.14.1](https://pypi.org/project/scipy/1.14.1/), [Astropy 6.1](https://pypi.org/project/astropy/6.1.0/), and [PyTables 3.10.1](https://pypi.org/project/tables/3.10.1/)), so compatibility would already be difficult.

It *might* have been possible to manually construct a set of dependencies that would have *worked* on <3.9, but I don't think that should be considered supported behavior, unless the pyproject.toml is modified to include it in the floor and there are unit tests; if a user really needs an old python version, they can just run an old COSMIC release that supported it; they would miss out on newer PRs, but it is too much developer effort to maintain support for python versions that are now formally considered [out of support](https://devguide.python.org/versions/). 

Additionally, having an unstated/untestsed best-effort version floor support can make future upgrades unexpectedly break that support; for example, right now I could run `pyupgrade` which currently flags 157 examples of legacy syntax in this repo that it can automatically or semi-automatically fix with more-or-less guaranteed safe fixes, but doing such upgrades increases the chance of breaking things. I considered doing this but I thought it was best left to a separate follow-up PR (which I can do if desired); there are several syntaxes in current COSMIC that may *break* in future versions if they are not upgraded eventually.

pyproject.toml
mcdigman: I can add the comment back but will wait for Katie to comment on the other issues

docs/pages/tutorials/adaptive/basics.rst
katiebreivik: sorry; let's go with `extras` here instead of `misc` or `merging-dco`

src/cosmic/utils.py
katiebreivik: I'm def in favor of dropping support for versions that are considered out of support by the Python dev guide!

pyproject.toml
katiebreivik: Yeah, I'd like to keep this around since we put that one in based on a previous headache [may be user/dev error on my part ha]

pyproject.toml
katiebreivik: let's change `merging-dco` to `extras` here and keep legwork. We can eventually add eventlab here as well once its a bit more means tested

katiebreivik (changes requested):
Thank you for opening this! Just a couple of requested changes in the comments below.

mcdigman:
@katiebreivik @TomWagg I renamed the dependency group in all 3 places and replaced the comment about bumping the numpy version. 

In the same vein, I added an additional comment to bump the dependency floor pins for the `lowest` CI job if they change in pyproject.toml. 

All CI is green except the one coverage job that *should* go green when it runs on push/future PRs after this PR merges.

katiebreivik (approved):
Thank you!

once we hear from @TomWagg , I'll go ahead with the merge!

TomWagg (approved):
Looks great to me, thanks @mcdigman!

